PR #21, opened 2018-01-16
add Landsat8 bands ratio

sshuair:
add landsat8 bands ratio in landsat8.py use `eval`.
and the ratio expression should be like this:

` '(B[5]-B[4])/(B[5]+B[4])'`,  B[5] means band 5,  B[4] means band 4

usage: 
```python
tile_z = 8
tile_x = 71
tile_y = 102
expression = '(B[5]-B[4])/(B[5]+B[4])'
result = landsat8.tile_ratio(LANDSAT_SCENE_C1, tile_x, tile_y, tile_z, expression)
```

### DIFF
--- a/rio_tiler/landsat8.py
+++ b/rio_tiler/landsat8.py
@@ -2,6 +2,7 @@
 
 from functools import partial
 from concurrent import futures
+import re
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


vincentsarago:
👋 @sshuair thanks for the PR. 
Code looks like 💯 (better than in https://github.com/mapbox/landsat-tiler/blob/master/app/landsat.py#L101 because it doesn't use `globals` 😱). 
That's said, I'd love to see if we could find out a way to create an utility function to work with `expression` for any file type (`L8`, `S2`, `CBERS`, html...). 

something like 
```python

from rio_tiler import main, landsat8, sentinel2, cbers
def expr(sceneid, expression, z, x, y, **kwarg):

# parse expression 

# get bands
if sceneid.startwith('L8'):
    tile = landsat8.tile(sceneid, x, y, z, bands)
elif sceneid.startwith('S2'):
    tile = sentinel2.tile(sceneid, x, y, z, bands)
elif sceneid.startwith('CB'):
    tile = cbers.tile(sceneid, x, y, z, bands)
else:
    # bands == band indexes 
    tile = main.tile(sceneid, x, y, z, rgb=bands)

# eval expression

return out
```

Also instead of thinking just for bands ratio I'd love if we could do `simple` expression like `[b4*0.2, b5*1.1, b2*0.9]` to create adjusted RGB image 

cc @sgillies

rio_tiler/landsat8.py
```diff
@@ -177,3 +178,89 @@ def tile(sceneid, tile_x, tile_y, tile_z, rgb=(4, 3, 2), tilesize=256, pan=False
...
+            'Tile {}/{}/{} is outside image bounds'.format(
+                tile_z, tile_x, tile_y))
+
+    mercator_tile = mercantile.Tile(x=tile_x, y=tile_y, z=tile_z)
+    tile_bounds = mercantile.xy_bounds(mercator_tile)
+
+    addresses = ['{}_B{}.TIF'.format(landsat_address, band) for band in bands]
+
+    _tiler = partial(utils.tile_band_worker, bounds=tile_bounds, tilesize=tilesize)
+    with futures.ThreadPoolExecutor(max_workers=3) as executor:
```
sgillies: Are you certain that multithreading improves performance here? Do you have benchmarks for it?

My understanding is that multithreading cannot improve the concurrency of CPU intensive work in Python. You really need to fork and use multiple processes. Threads in Python are useful for I/O operations that otherwise block for small amounts of time. See Chapter 5 in Brett Slatkin's Effective Python book.

sshuair:
@vincentsarago @sgillies thanks for your comment and suggestion. I'll try `snuggs` or `numexpr` in later coding.

And it seems that @vincentsarago has finished the bands ratio using a better way according to #22 👏 👍

vincentsarago:
closing in favor of #22

PR #22, opened 2018-01-17
[WIP] add expression utily function

vincentsarago:
ref #21 

cc @sshuair


sshuair:
Looks very good. is there anything else I can help with?

PR #27, opened 2018-02-15
rm read() from setup.py

samsammurphy:
It doesn't look the `read()` method is being used, unless maybe it is being imported by another module somewhere?


vincentsarago:
good catch ;-)

PR #28, opened 2018-02-15
updated comments in band_min_max_worker()

samsammurphy:
This method seems abstract enough to handle any image band URL.


vincentsarago:
thanks @samsammurphy :-)

PR #32, opened 2018-03-14
bytescale

samsammurphy:
Casting an image to UINT8 might produce surprising results if not rescaled to between 0 and 255. Scipy have a neat function to handle this called bytescale(). It also allows users to define min and max values for the rescale (e.g. the percentiles returned by the metadata() methods).

I would suggest using it (or something like it). The example here has copied and pasted bytescale() into rio_tiler.utils and updated array_to_img().

The effect is that, if people display or save the Pillow image created by array_to_img() it _should_ always show something pretty.

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -411,8 +411,39 @@ def cbers_parse_scene_id(sceneid):
 
     return meta
 
+def bytescale(arr, cmin=None, cmax=None, high=255, low=0):
+    """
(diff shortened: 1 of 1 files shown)


vincentsarago:
👀 I'm not sure to see how `bytescale` will be different from https://github.com/mapbox/rio-tiler/blob/master/rio_tiler/utils.py#L182 ?

samsammurphy:
Gosh! Apologies for missing that. Yep, looks like are essentially the same.

PR #34, opened 2018-03-26
rename rgb option to indexes and bands

### LINKED ISSUE #33: rename `rgb` option to `indexes` or `bands`
in `.main` we use `rgb` option to pass band indexes. If we want to use the same terminology as in rasterio we should use `indexes`. 

For `.sentinel2`, `.landsat8` and `cbers` I feel we should use the word `bands` (!= indexes, because we read individual band files) instead of `rgb`. 

cc @samsammurphy

vincentsarago:
closes #33

### DIFF
--- a/rio_tiler/main.py
+++ b/rio_tiler/main.py
@@ -9,11 +9,11 @@ from rio_tiler.errors import TileOutsideBounds
 
 
 def bounds(address):
-    """Retrieve image bounds.
+    """
+    Retrieve image bounds.
 
     Attributes
     ----------
-
     address : str
         file url.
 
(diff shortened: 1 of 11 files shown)


rio_tiler/main.py
```diff
@@ -44,32 +44,32 @@ def tile(address, tile_x, tile_y, tile_z, rgb=None, tilesize=256, nodata=None, a
...
     -------
     data : numpy ndarray
     mask: numpy array
-    """
 
+    """
     with rasterio.open(address) as src:
         wgs_bounds = transform_bounds(
             *[src.crs, 'epsg:4326'] + list(src.bounds), densify_pts=21)
-        nodata = nodata if nodata is not None else src.nodata
```
vincentsarago: Not mentioned but, this a `big` change (more a bug fix) we should let `gdal/rasterio` handle internal nodata value (read_masks).

PR #58, opened 2019-01-08
add Landsat pan band (08) values in histogramm min/max fetch

vincentsarago:
This PR adds Landsat8 Panchromatic min/max fetch when getting Landsat metadata.

cc @sgillies @jacquestardie

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,7 @@
+Next
+----
+- fetch Landsat panchromatic band min/max (#58)
+
 1.0rc2 (2018-08-22)
 -------------------
 - add test case for pix4d nodata+alpha band data
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 Next
 ----
-- fetch Landsat panchromatic band min/max (#58)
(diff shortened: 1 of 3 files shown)


rio_tiler/landsat8.py
```diff
@@ -70,7 +70,7 @@ def metadata(sceneid, pmin=2, pmax=98):
     info = {'sceneid': sceneid}
     info['bounds'] = toa_utils._get_bounds_from_metadata(meta_data['PRODUCT_METADATA'])
 
-    bands = ['1', '2', '3', '4', '5', '6', '7', '9', '10', '11']
+    bands = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11']
```
sgillies: @vincentsarago will you add a note about this to the change log?

vincentsarago:
thanks @sgillies

CHANGES.txt
```diff
@@ -1,3 +1,7 @@
+Next
+----
+- fetch Landsat panchromatic band min/max (#58)
+
```
sgillies: The bug fix here is that band 8 was missing, yes? Let's say so in the change log.

Sorry to be a pain!

vincentsarago: I'll re-phrase

PR #73, opened 2019-02-05
1.0rc3

vincentsarago:
PR for official version `1.0.0`


nrweir:
Thanks for the heads up @vincentsarago!

mojodna:
Awesome! Thanks for the heads up as well.

marblecutter only uses `utils.linear_rescale` from rio-tiler at the moment, so don't worry about me with the breaking changes.

PR #97, opened 2019-04-01
first sketch of mercator utility functions

### LINKED ISSUE #96: add zoom range and resolution in metadata output
Having those could be useful IMO

vincentsarago:
resolves #96 

TO DO
- [x] tests
- [x] validate math
- [x] add tests
- [x] write docs 

cc @DanSchoppe


rio_tiler/mercator.py
```diff
@@ -0,0 +1,109 @@
...
+
+    max_zoom = zoom_for_pixelsize(corrected_resolution, tilesize=tilesize)
+
+    ovr_resolution = corrected_resolution * max(h, w) / tilesize
+    min_zoom = zoom_for_pixelsize(ovr_resolution, tilesize=tilesize)
+
+    return (min_zoom, max_zoom)
+
+
+def get_mercator_info(src_dst, latitude_correction=False, tilesize=256):
```
vincentsarago: > latitude_correction=False

We need to change the semantic

Issue #56, opened 2018-11-28
Explore DHI-GRAS/terracotta and bring change over rio-tiler 


vincentsarago:
https://github.com/DHI-GRAS/terracotta is a new lib inspired by rio-tiler made by @dionhaefner et all. 
This lib has plenty of nice tricks which could be useful in rio-tiler
eg:
- different resampling method for up and down sampling: https://github.com/DHI-GRAS/terracotta/blob/7aa0751a105d1a804f0c5389e0abaab4399ffdee/terracotta/drivers/raster_base.py#L431-L436
- check tile coverage https://github.com/DHI-GRAS/terracotta/blob/7aa0751a105d1a804f0c5389e0abaab4399ffdee/terracotta/drivers/raster_base.py#L428-L429
- prevent interpolation artefact ? https://github.com/DHI-GRAS/terracotta/blob/7aa0751a105d1a804f0c5389e0abaab4399ffdee/terracotta/drivers/raster_base.py#L396-L411

dionhaefner:
> prevent interpolation artefact

Maybe I can give you some background on this one. I found that `rio-tiler` has some trouble with visible tile edges at high zoom levels if `resampling` was set to anything else than `nearest`.

If I understand the problem correctly, this occurs because the VRT ends right at the tile boundary, so the entity reading from it cannot know what lies beyond, and how it should interpolate. Padding the VRT with 2 pixels in each direction fixes the problem (if you're careful with roundoff errors). It has to be two pixels because `cubic` interpolation has a bandwidth of 2, I guess.

dionhaefner:
You might also like this re-implementation of `_calculate_default_transform`:

https://github.com/DHI-GRAS/terracotta/blob/7aa0751a105d1a804f0c5389e0abaab4399ffdee/terracotta/drivers/raster_base.py#L302

I found it to perform unconditionally better (i.e., more accurate) when round-tripping reprojections, especially at high latitudes (we had some imagery from northern Greenland). The problem is that GDAL's `_calculate_default_transform` enforces square pixels in the target CRS, which destroys information if the image gets elongated along one dimension.

vincentsarago:
@dionhaefner 
[...]

PR #104, opened 2019-05-04
Provide parameter for retrieving additional data around tile edges to minimise edge effects

rowanwins:
As [proposed](https://github.com/cogeotiff/rio-tiler/issues/56#issuecomment-488920779) in #56 this PR aims to reduce the edge effects associated with tiles by introducing a `tile_edges_padding` parameter. Additional data is retrieved around the tiles edges during the tile generation process which helps to reduce sharp edges that occur due to resampling.

@vincentsarago did suggest
> and then overwrite it to 0 if we don't need resampling (if the internal tiles match the requested bounds)

I'm not really sure how we'd know about the internal tiles so happy to take any advice on this one :)

Let me know if you need anything else on this.


vincentsarago:
thanks @rowanwins this is great. I need to think about how to check if the tile is aligned then also check the performances (which should not vary).

rio_tiler/utils.py
```diff
@@ -298,6 +303,18 @@ def _tile_read(
     )
 
     vrt_transform, vrt_width, vrt_height = get_vrt_transform(src_dst, bounds)
+    out_window = None
+    if tile_edges_padding > 0:
+        vrt_transform = vrt_transform * Affine.translation(-tile_edges_padding, -tile_edges_padding)
+        orig__vrt_height = vrt_height
+        orig_vrt_width = vrt_width
+        vrt_height = vrt_height + 2 * tile_edges_padding
```
vincentsarago: I'm ok with having edges being a multiple of 2 but not sure if this really needed ? also if we keep it like that the doc string should mention it.

vincentsarago:
@rowanwins maybe we can do something like 

```python
def is_aligned(src_dst, bounds, tilesize):
    """Check if tile is aligned with internal tiles."""
    if src_dst.crs != CRS.from_epsg(3857):
        return False

    col_off, row_off, w, h = windows.from_bounds(
        *bounds, height=tilesize, transform=src_dst.transform, width=tilesize
    ).flatten()

    if round(w) % 64 and round(h) % 64:
        return False

    if (src_dst.width - round(col_off)) % 64:
        return False

    if (src_dst.height - round(row_off)) % 64:
        return False

    return True
```
The idea is to check: 
- if the CRS is EPSG:3857
- if the size of the window is a multiple of 64 px (minimum size of internal tile) 
- if the offset is a multiple of 64px

### why 64px ? 
ideally we want to check if the window is a multiple of the internal tile size (raw data and overview) but we don't have a way to do it with rasterio and I'm not sure this is needed. By checking if we have a multiple of 64 we make sure that even on border tiles we don't need to do resampling.

aligned_tile

Note: this might be totally wrong, but did a quick check with a web-optimized file and it work

rio_tiler/utils.py
```diff
@@ -298,6 +303,18 @@ def _tile_read(
     )
 
     vrt_transform, vrt_width, vrt_height = get_vrt_transform(src_dst, bounds)
+    out_window = None
```
vincentsarago: also here can we set `out_window` to `out_window = windows.Window(col_off=0, row_off=0, width=vrt_width, height=vrt_height)`

rio_tiler/utils.py
rowanwins: I feel like the padding value makes most sense as a single edge so I'll update the docstring to make that clearer.

rowanwins:
Hi @vincentsarago 

I've made those changes. 
I think it would be good to add a test around the `_requested_tile_aligned_with_internal_tile` however I wasn't sure if any of the test fixtures had those internal web-optimised tiles, if you can point me in the right direction there that would be great.

vincentsarago:
@rowanwins you can use this file https://www.dropbox.com/s/op0szbw18szqupc/web.tif?dl=0

tests/test_utils.py
```diff
@@ -80,6 +81,53 @@ def test_tile_read_validResampling():
     assert mask.shape == (16, 16)
 
 
+def test_resampling_returns_different_results():
```
vincentsarago: 👍

tests/test_utils.py
```diff
@@ -80,6 +81,53 @@ def test_tile_read_validResampling():
...
+        3835304.331237001,
+    )
+    tilesize = 16
+
+    arr, mask = utils.tile_read(address, bounds, tilesize)
+    arr2, mask2 = utils.tile_read(address, bounds, tilesize, tile_edge_padding=0)
+    assert not np.array_equal(arr, arr2)
+
+
+def test_tile_padding_only_effects_edge_pixels():
```
vincentsarago: 👍

rowanwins:
Hi @vincentsarago 

I attempted to change the base but I'm not sure if that worked TBH (I ran `git rebase --onto maint-1.0` followed by `git push -force origin pad-tile`), my git skills are average :)

I think everything else is ticked off (other than codecov complaining about a decrease in coverage). It would be good to test that `_requested_tile_aligned_with_internal_tile` function a little more, I'm just not entirely sure how...

vincentsarago:
this is great @rowanwins I'll take it from here. thanks for your help

rowanwins:
Out of interest how did you change the base branch @vincentsarago ?

PR #90, opened 2019-03-11
Add rplumbo colormap

DanSchoppe:
In remote sensing, NDVI (measure of vegetation) is frequently colored red (bad) to green (good). (Sources: [0](https://pro.arcgis.com/en/pro-app/help/data/imagery/ndvi-colorized-function.htm), [1](https://crops.extension.iastate.edu/cropnews/2016/05/choosing-right-imagery-best-management-practices-color-nir-and-ndvi-imagery) )

Unfortunately, there are no red-to-green color maps supported by rio-tiler. This PR adds the "rplumbo" color map:
- Grayscale (dark-to-light) from indices 0-127
- Red-to-green from indices 128-255
Screen Shot 2019-03-11 at 9 44 40 AM

An aside: In the long-term, it would be nice if rio-tiler could take a color map (or some representation that programmatically generates a color map) via query parameters or even POST body. Is this something worth adding an issue for?

Thanks to Mapbox and all contributors for this amazing rasterio / rio-tiler toolchain!

## References:
[0] https://pro.arcgis.com/en/pro-app/help/data/imagery/ndvi-colorized-function.htm
```
The default color map is named NDVI3. Red and orange pixels represent areas with no vegetation. Yellow pixels represent areas with low to moderate vegetation. Green pixels represent areas with high vegetation density and vigor.
```

[1] https://crops.extension.iastate.edu/cropnews/2016/05/choosing-right-imagery-best-management-practices-color-nir-and-ndvi-imagery
```
Areas with Green had the highest index value, followed by Yellow and Red had the lowest.
```

PR #110, opened 2019-06-19
Add grayscale colormap

DanSchoppe:
Similar to #90, this PR adds a grayscale colormap to the project (along with a relevant unit test).


DanSchoppe:
@vincentsarago I'm not aware of the reasons for using `cogeotiff:maint-1.0`. Perhaps it's to aggregate a number of improvements in order to reduce the number of releases? Would I have a better chance at getting this into a release by PRing into `master`?

vincentsarago:
@DanSchoppe thanks for this PR I love it.

Can you be more explicit about this PR and what it solve ? To be honest I'm not sure to fully get why it is usefull.

> I'm not aware of the reasons for using cogeotiff:maint-1.0. Perhaps it's to aggregate a number of improvements in order to reduce the number of releases? Would I have a better chance at getting this into a release by PRing into master?

Yeah the idea was to always PR on maint-1.0 (as rasterio does) and then when there is enough or important features/fix we can merge it to master and release.

DanSchoppe:
@vincentsarago Thanks for the PR feedback. I apologize, I should have drafted an explanation in the PR description. Here's what I'm thinking:

I've got a number of float32 COGs that contain scientific values. Since tile image formats (JPG, PNG) do not support floating-point values, the only way to generate a tile containing these floating point values is to apply a reversible (although unfortunately lossy) colormap to the data.

For example, a COG with float32 values ranging from [-1, 1]. A web client requests a tile with grayscale colorization. The resulting tile has uint8 bands with values in the range [0, 255]. The web client can then re-map the [0, 255] range to the original [-1, 1] COG range to recover a (lossy) estimate of the original floating point value.

Does that make sense?

vincentsarago:
> For example, a COG with float32 values ranging from [-1, 1]. A web client requests a tile with grayscale colorization. The resulting tile has uint8 bands with values in the range [0, 255]. The web client can then re-map the [0, 255] range to the original [-1, 1] COG range to recover a (lossy) estimate of the original floating point data.

> Does that make sense?

👋 @DanSchoppe, still I'm not fully understanding. 

If you have float data from -1 to 1, to enable the `image` creation you'll need to `rescale` it to 0-255 value (eg by using https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L385-L408) so the output value is already in a `grayscale` colormap like.

DanSchoppe:
@vincentsarago Thanks for the feedback.

I think you're right. I didn't realize I could rescale the image values without providing a colormap. This PR was essentially implementing a `null` colormap, but I see now that this is unnecessary.

vincentsarago:
Excellent @DanSchoppe 

>  I didn't realize I could rescale the image values without providing a colormap

I should maybe add a note in the documentation, but yes `colormap` are just there to create a map to represent 1 band range value from 0-255 to a 3 band RGB image (in https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L491-L563). Rescaling as to be done before passing the tile array to the `array_to_image` function

PR #92, opened 2019-03-12
Allow custom discrete colormap (ref: #91)

vincentsarago:
ref #91 

This is a first sketch of dictionary support for color maps

# TO DO 
- [x] benchmark 
- [x] update tests 
- [ ] coms 

This is a breaking change, I'm kinda hesitant to make a 2.0 release for this change but more inclined  to make a 1.2.0 release if we go forward with this.

Note: This change will enable custom discrete colormap like: 
```

        colormap = {
            322: [209, 46, 198],
            324: [51, 149, 229],
            328: [116, 193, 214],
            336: [132, 240, 110],
            352: [178, 204, 31],
            368: [29, 88, 215],
            386: [31, 211, 43],
            388: [119, 214, 192],
            392: [104, 226, 28],
            400: [228, 57, 100],
            416: [23, 40, 230],
            432: [237, 66, 14],
            480: [198, 89, 219],
            834: [201, 24, 148],
            836: [121, 210, 169],
            840: [230, 217, 73],
            848: [144, 106, 210],
            864: [212, 47, 127],
            880: [170, 119, 203],
            898: [204, 172, 103],
            900: [54, 28, 204],
            904: [210, 63, 63],
            912: [218, 147, 84],
            928: [55, 208, 208],
            944: [196, 239, 126],
            992: [101, 205, 132]
        }
```

cc @DanSchoppe

PR #115, opened 2019-07-16
add more colormap and switch to numpy binaries

### LINKED ISSUE #91: More colormap options!
With #90 @DanSchoppe added a third `colormap` to rio-tiler. While this is awesome, I think rio-tiler could do more for allowing user to define custom colormap.

**colormap** are used when translating a one band data array to the image format using the new `array_to_image`(using GDAL using a simple translation:
https://github.com/cogeotiff/rio-tiler/blob/e512d8441e7ebb460d074a4a98a76051ad983b73/rio_tiler/utils.py#L386

So here ☝️ a `rio_tiler.utils.array_to_image` compatible colormap object is usually an array of shape `(256, 3)`
```python
c = get_colormap(format="gdal")
print(c.shape)
 (256, 3)
print(c)
[
  [255 255 255]
  [250 250 250]
  [246 246 246]
  [242 242 242]
  ...
  [255   1 191]
  [255   1 207]
  [255   1 223]
  [255   1 239]
]
```

I first though we could have used a custom dict like
```python
{
    0: [255, 255, 255],
    100: [255,   1, 191],
    ...
}
```
~to create a colormap for`discrete` values, but dict are not hashable so the current code in `rio_tiler.utils.array_to_image` won't work.~

I do love the dict option for colormap (like Rasterio use https://github.com/mapbox/rasterio/blob/960a906dad2a4e426387ce048a52c6e90afdcd2b/docs/topics/color.rst#writing-colormaps) so I first check if we can maybe change the `rio_tiler.utils.get_colormap` function to have to return a dictionary colormap (because it seems to be the default format for GDAL) and in the same time adapt the `rio_tiler.utils.array_to_image` to work with dict. 

In a second time we could add another utility function to create colormaps, it seems that matplotlib has some functionality but I'm not sure we should had more required module to rio-tiler.

vincentsarago:
closes #91 

after https://github.com/cogeotiff/rio-tiler/pull/92, this PR refactor how we store color map (.txt to numpy binary .npy) 

https://github.com/cogeotiff/rio-tiler/blob/5ea11e008b840d8b2a41f7f96f0540d73bc26dca/rio_tiler/cmap/__init__.py#L2-L19

@DanSchoppe could you please make sure the `rplumbo` colormap still work for you ? 

@dionhaefner thanks for your help on this one 🙏


DanSchoppe:
Thanks for this! Investigating now...

vincentsarago:
@DanSchoppe I'm going to merge this today except if you find something ;-)

DanSchoppe:
Hey @vincentsarago, sorry I didn't get positive confirmation on this before you needed to merge.

I took a quick crack at testing this yesterday but was getting errors when calling `utils.array_to_image()` with a colormap. Turns out I had forgotten about the `format="gdal"` param in the call to `utils.get_colormap()` 🙄

Anyway, I was able to get this working and it looks good! No diff on the tile returned from a tile retrieved before/after this patch.

PR #119, opened 2019-09-18
use same overview size calculation as in GDAL

### LINKED ISSUE #116: Possibly incorrect calculation of overview size
Hi! Thanks for a really great library, it has been a breeze working with, and the code is very clear to read.

While working with rio-tiler I stumbled upon what I think might be an error in how the size of  overviews are calculated in `utils.raster_get_stats` (https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L187): since integer division is used, any fractions will always be rounded down. Looking at the output from `gdalinfo`, it seems to always round fractions _up_ to the next integer size instead. I have not been able to find out what the spec actually says, but I am tempted to trust `gdalinfo`.

If my guess is correct, this might cause the overview to be unnecessarily scaled since the size is off by one.

vincentsarago:
closes #116 

This PR do: 
- use `_div_round_up` to make use we use the same overview size as in GDAL
- add `resampling_method` in `raster_get_stats`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,7 +1,8 @@
 
 NEXT (TBD)
 ----------
-- reduce memory footprint of expression tiler (#18d4acc)
+- reduce memory footprint of expression tiler
+- fix wrong calculation for overview size in `raster_get_stats` (#116)
 
 1.2.10 (2019-07-18)
 -------------------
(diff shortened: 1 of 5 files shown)


vincentsarago:
output values for `raster_get_stats` might change slightly because of the change in the resampling method

PR #126, opened 2019-11-05
Fixed PNG bands count

pierotofy:
Hello :hand:

This PR fixes an issue I've encountered while implementing a Django tiler using rio-tiler. When serving PNG tiles with alpha, the band count is wrongly set to `5`?

The following changes allow the tiles to load properly.

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -620,7 +620,7 @@ def array_to_image(
     if img_format == "webp" and arr.shape[0] == 1:
         arr = np.repeat(arr, 3, axis=0)
 
-    if mask is not None and img_format != "jpeg":
+    if mask is not None and not img_format in ["jpeg", "png"]:
         nbands = arr.shape[0] + 1
     else:
         nbands = arr.shape[0]
@@ -639,7 +639,7 @@ def array_to_image(
             dst.write(arr, indexes=list(range(1, arr.shape[0] + 1)))
 
             # Use Mask as an alpha band
-            if mask is not None and img_format != "jpeg":
+            if mask is not None and not img_format in ["jpeg", "png"]:
                 dst.write(mask.astype(arr.dtype), indexes=nbands)
 
         return memfile.read()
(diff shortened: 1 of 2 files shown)


vincentsarago:
thanks for this PR @pierotofy, 
Can you add a test to test this change ? 

thanks

pierotofy:
Yes I'll add test cases as necessary. :+1:

This happens when a RGBA image with a nodata mask is tiled with the PNG driver. The `nbands = arr.shape[0] + 1` adds an extra alpha band (bringing the total to `5`).

pierotofy:
@vincentsarago please let me know if this looks OK or if further changes are needed. Thank you!

vincentsarago:
@pierotofy, now I understand thanks for your update.

I'm a little worry about the changes. Basically this tries to overcome some user problems, meaning it's up to the user to pass valid data to the `array_to_image` to image function. In your case, because the user wants a PNG, the inputs should be a 4bands RGBA array or a 3 bands RGB array + a mask but passing a 4bands RGBA + a mask doesn't really make sense to me. 

if you're input dataset is an RGBA raster, I think the correct way to work with that is to use `indexes=(1, 2, 3)` to return a RGB array + a mask 

I've already added some `tests` to cover corner case about jpeg and 1band webp and I'd like to avoid more `if driver` case.

pierotofy:
Ah, makes sense. I can simply avoid the mask if the band count is 4. Note you might have a bug in https://github.com/RemotePixel/remotepixel-tiler/blob/master/remotepixel_tiler/cogeo.py#L157 then :)

```
>>> from rio_tiler import main
>>> r, m = main.tile('./odm_orthophoto.tif', 1025342, 1477931, 22)
>>> r.shape
(4, 256, 256)
>>> m.shape
(256, 256)
```

orthophoto file: https://drive.google.com/open?id=1AIbojXYELmecjgEN8mgkmmm7NIrXsVtf

I'll close this one. Thank you for your help!

vincentsarago:
Thanks @pierotofy. 

I'll open a new issue to see if we want to remove the alpha band from the output arr when `indexes` is not provided.

PR #134, opened 2019-12-06
Expr stats

pierotofy:
Per https://github.com/OpenDroneMap/rio-tiler/pull/1

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -91,6 +91,7 @@ def _raster_get_stats(
     histogram_range=None,
     resampling_method="bilinear",
     warp_vrt_option={},
+    expr=None,
 ):
     """
     Retrieve dataset statistics.
@@ -122,7 +123,9 @@ def _raster_get_stats(
         Resampling algorithm.
     warp_vrt_option: dict, optional (default: {})
         These will be passed to the rasterio.warp.WarpedVRT class.
-
+    expr : str, optional (default: None)
+        Expression to apply to raster before computing statistics
+        
     Returns
     -------
     out : dict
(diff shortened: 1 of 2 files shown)


vincentsarago:
thanks for the PR @pierotofy, I'll make some comment and then we can add some tests :-)

rio_tiler/utils.py
```diff
@@ -218,6 +221,21 @@ def _get_descr(ix):
             masked=True,
         )
 
+        if expr is not None:
+            bands_names = tuple(set(re.findall(r"(?P<bands>b[0-9A]{1,2})", expr)))
```
vincentsarago: we can remove the `A` for this because the expression should only correspond to band indexes. Also maybe the limitation to 2 number might not be necessary

pierotofy:
Hey @vincentsarago :hand: sorry for the delayed response. Busy week ahead, I'll try to find some time here and there to make the necessary changes and add tests.

rio_tiler/utils.py
pierotofy: Something like `(?P<bands>b[0-9]+)`?

pierotofy:
Modified regex, added basic test case. Not sure if it's sufficient or more test coverage should be added. :+1:

vincentsarago:
thanks the the update @pierotofy .

I'm having a bit of a second though on this, what happens if someone pass indexes and expr ? 

also on the returned dict, we don't see the diff between expr and indexes ... I'm sorry if this add more complexity. 

what we could simply do is:
- not allow indexes + expr, 
- get stats for all band (because no indexes) + expr 
- add `expression: {rgb[1]: stats, rgb[2]: stats ... }` in the returned dictionnary

pierotofy:
Mm, yes there are scenarios that would cause troubles. How about a simple:

```
# pseudo code
if indexes and expr:
    raise "not allowed"
```

I'm not sure I understand the other two points. Like I mentioned earlier, perhaps this is a feature too specific to WebODM (where we handle only a subset of possible inputs and we don't use indexes), so no problem if this can't be included.

vincentsarago:
I'm going to close this. 

With the new `COGReader` it's a bit easier to achieve this.


```python
with COGReader("my.tif") as cog:
    info = cog.info.copy()
    info["statistics"] = cog.stats(expression="b1/b2")
    return info
```

PR #138, opened 2020-02-01
add benchmarks

vincentsarago:
This PR adds a benchmarking suite.

### To Do
- run bencmarks on circle-ci (would be great to run on gdal 2.4 and gdal 3.0)
- compare run with released

### Question
- what do we do with the report ? (save to s3 ?)
- is the a way to block a PR/release if performance decrease ?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,20 @@ jobs:
         environment:
           - TOXENV=py37
 
+  benchmark:
+    docker:
+      - image: circleci/python:3.7.2
+    working_directory: ~/rio-tiler
+    steps:
+     - checkout
+     - run: pip install -e .[test] --user
+     - run: |
+         python -m pytest \
+         --benchmark-only \
+         --benchmark-autosave \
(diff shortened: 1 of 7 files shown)


vincentsarago:
Note: I've created some rasterio+gdal docker image to be able to run benchmark on multiple version of GDAL 
https://hub.docker.com/repository/registry-1.docker.io/remotepixel/alpine-rasterio

vincentsarago:
FYI, I'm still figuring out what to do with the benchmark results but here a first result:

```
Reading boundless tile for a uint8 + internal mask
---------------------------------------------------------------------------
Name (time in ms)    Min          Max                Mean            Median
---------------------------------------------------------------------------
GDAL2.4          83.7908      94.4726       87.5320 (1.0)     87.0866 (1.0)  
GDAL3.0         122.6241     131.7297     125.5178 (1.43)   124.5531 (1.43) 
GDALmaster      140.0582     371.1401     256.1110 (2.93)   227.2926 (2.61) 
```
I guess this show a point why benchmark is important ;-)

PR #141, opened 2020-02-14
[Bug] Remove band duplicates while preserving order

pierotofy:
Found this while troubleshooting a related problem. The `tuple(set(x))` expression is used to avoid duplicate band names, however, since the order of the bands is important in the subsequent `zip` call around https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L779 this can cause impredictable behavior, since the ordering of `tuple(set(x))` is not guaranteed:

```python
a = ['c', 'a', 'b']
print("tuple(set(a)): %s" % str(tuple(set(a))))
print("tuple(sorted(set(a))): %s" % str(tuple(sorted(set(a)))))
```

```bash
# python --version
Python 3.6.9

# python tmp/test.py 
tuple(set(a)): ('a', 'b', 'c')
tuple(sorted(set(a))): ('a', 'b', 'c')

# python tmp/test.py 
tuple(set(a)): ('b', 'a', 'c')
tuple(sorted(set(a))): ('a', 'b', 'c')
```

This PR addresses this problem.

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -748,7 +748,7 @@ def expression(sceneid, tile_x, tile_y, tile_z, expr=None, **kwargs):
     if not expr:
         raise Exception("Missing expression")
 
-    bands_names = remove_duplicates(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr))
+    bands_names = tuple(sorted(set(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr))))
     rgb = expr.split(",")
 
     if sceneid.startswith("L"):
@@ -803,12 +803,3 @@ def pansharpening_brovey(rgb, pan, weight, pan_dtype):
         ratio = _calculateRatio(rgb, pan, weight)
         return np.clip(ratio * rgb, 0, np.iinfo(pan_dtype).max).astype(pan_dtype)
 
-
-def remove_duplicates(seq):
-    """
-    Remove duplicates from a sequence while preserving order.
-    Adapted from https://stackoverflow.com/a/480227
-    """
-    seen = set()
-    seen_add = seen.add
-    return tuple([x for x in seq if not (x in seen or seen_add(x))])
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -748,7 +748,7 @@ def expression(sceneid, tile_x, tile_y, tile_z, expr=None, **kwargs):
     if not expr:
         raise Exception("Missing expression")
 
-    bands_names = tuple(set(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr)))
+    bands_names = remove_duplicates(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr))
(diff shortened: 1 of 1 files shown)


vincentsarago:
@pierotofy there is no obvious problem for me here ...

let's look at the code
https://github.com/cogeotiff/rio-tiler/blob/5b57eb0a4c09c499f79bf252ff2ca196e9ef7294/rio_tiler/utils.py#L751-L786


```python
    # Example: expr = "b1,b3,b2,b1/b3"
    bands_names = tuple(set(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr)))
    # bands_names = ('2', '3', '1')

    rgb = expr.split(",")
    # rbg = ['b1', 'b3', 'b2', 'b1/b3']
    
    ....

    bands_names = ["b{}".format(b) for b in bands_names]
    # bands_names = ['b2', 'b3', 'b1']
    
    # the tile got bands = ("2", "3", "1") so I believe
    # it should return a (3, 256, 256) array with bands in order of ("2", "3", "1") 
    
    arr = dict(zip(bands_names, arr))
    # Here we create a dict with
    {
        "b2": (1, 256, 256)
        "b3": (1, 256, 256)
        "b1": (1, 256, 256)
    }
```

I may be totally wrong. Do you have an example where ☝️ is failling

pierotofy:
```python
import re
import numpy as np

# Example: expr = "b1,b3,b2,b1/b3"
expr="b1,b3,b2,b1/b3"
bands_names = tuple(set(re.findall(r"b(?P<bands>[0-9A]{1,2})", expr)))
print("EXPECTED: bands_names = ('2', '3', '1')")
print(bands_names)

rgb = expr.split(",")
print("EXPECTED: rbg = ['b1', 'b3', 'b2', 'b1/b3']")
print(rgb)

bands_names = ["b{}".format(b) for b in bands_names]
print("EXPECTED: bands_names = ['b2', 'b3', 'b1']")
print(bands_names)
```

```bash
EXPECTED: bands_names = ('2', '3', '1')
('3', '2', '1')
EXPECTED: rbg = ['b1', 'b3', 'b2', 'b1/b3']
['b1', 'b3', 'b2', 'b1/b3']
EXPECTED: bands_names = ['b2', 'b3', 'b1']
['b3', 'b2', 'b1']
```

Might take a few runs of the script to trigger the results.

vincentsarago:
my point is I don't see how this influence the result in from numexpr 🤔 

from what I understand the order of `bands` won't have any impact

pierotofy:
If it doesn't impact the results, then you can close this; I was troubleshooting a related problem, but thought the nondeterministic nature of this might cause issues in other areas.

:clinking_glasses:

vincentsarago:
let me know how the WebODM problem goes, maybe it's on rio-tiler side but for now I'm not seeing it :-(

pierotofy:
You're right, it doesn't affect the results due to the indexes= parameter. Issue on WebODM is solved ( wasn't due to this one).

PR #142, opened 2020-02-17
Mask resampling + binary mask option

### LINKED ISSUE #105: Apply same resampling method for mask and tile
in https://github.com/cogeotiff/rio-tiler/blob/2bf9f144958ae226af4db338487e8a34159550be/rio_tiler/utils.py#L345-L353
we could end up (actual behavior) with using different resampling for mask and data which will result in bad masking.

vincentsarago:
This is a second try to fix #105 and also makes alpha/mask band read more consistent for all dataset.

This is a breaking change and will need to be tested before release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,4 +105,4 @@ ENV/
 
 .benchmarks/
 tests/benchmarks/data/*
-tests/fixtures/mask/*.tif
+tests/fixtures/mask*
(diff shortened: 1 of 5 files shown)


vincentsarago:
Note: benchmarks show a slight speed decrease (~10ms) but my guess it that bilinear resampling is a bit slower then nearest.

vincentsarago:
this now lives in #154

PR #143, opened 2020-02-18
adapt for new sentinel L2A level

### LINKED ISSUE #137: Sentinel-2 Level-2A Bottom of Atmosphere Imagery
Hello 👋! ESA has globally rolled out bottom of atmosphere "Level-2A" (aka "L2A") corrected imagery as of December 2018. Are there any known technical limitations for supporting this imagery? If not, can we extend this project to support L2A?

At first glance, it looks like L1C and L2A datasets adhere to a similar (although not identical) S3 schema:
```
$ aws s3api list-objects --bucket sentinel-s2-l1c --prefix tiles/1/C/EM/2019/11/20/0 --request-payer requester
{
    "Contents": [
        {
            "Key": "tiles/1/C/EM/2019/11/20/0/B01.jp2",
            "LastModified": "2019-11-20T19:55:25.000Z",
            "ETag": "\"96effcc32e2c0f0d6cc45bf18efbb8ce\"",
            "Size": 297678,
            "StorageClass": "INTELLIGENT_TIERING",
            "Owner": {
                "ID": "91d380b3cead28df927c824731b0173701336cd8d67b0679d7166288f3850f38"
            }
        },
        ...
```

```
$ aws s3api list-objects --bucket sentinel-s2-l2a --prefix tiles/1/C/EM/2019/11/20/0 --request-payer requester
{
    "Contents": [
        {
            "Key": "tiles/1/C/EM/2019/11/20/0/R60m/B01.jp2",
            "LastModified": "2019-11-20T21:10:52.000Z",
            "ETag": "\"620a1f03d96f1df8bdfdd32b64c90608\"",
            "Size": 314419,
            "StorageClass": "INTELLIGENT_TIERING",
            "Owner": {
                "ID": "91d380b3cead28df927c824731b0173701336cd8d67b0679d7166288f3850f38"
            }
        },
        ...
```

Notice the addition of the "R60m" to the prefix of the L2A dataset. This seems to indicate the ground sample distance (GSD) of the dataset. Bands with finer resolution have a variety of GSD available. For example, B02 has `R10m/B02.jp2`, `R20m/B02.jp2`, and`R60m/B02.jp2` variants.

It looks like rio-tiler hard-codes top of atmosphere ("L1C") support:
https://github.com/cogeotiff/rio-tiler/blob/6b3b57e4d837cfa855d9b4a74175812339ac433e/rio_tiler/sentinel2.py#L19
I would imagine L2A support being added behind an optional parameter to `bounds` / `metadata` / `tiles` functions?

vincentsarago:
closes #137 

With this we still support old scene id but print a warning!


rio_tiler/sentinel2.py
```diff
@@ -100,34 +142,67 @@ def _sentinel_parse_scene_id(sceneid):
...
+    if band in SENTINEL_L2_BANDS["60m"]:
+        for res, bands in SENTINEL_L2_BANDS.items():
+            if band in bands:
+                return "R{}/B{}".format(res, band)
+    elif band in SENTINEL_L2_PRODUCTS["60m"]:
+        for res, bands in SENTINEL_L2_PRODUCTS.items():
+            if band in bands:
+                return "R{}/{}".format(res, band)
+    else:
+        raise InvalidBandName("{} is not a valid Sentinel band name".format(band))
```
vincentsarago: This is where we choose the highest resolution for each L2A band

rio_tiler/sentinel2.py
```diff
@@ -100,34 +142,67 @@ def _sentinel_parse_scene_id(sceneid):
...
+        meta["bands"] = SENTINEL_L2_BANDS["60m"]
+        meta["valid_bands"] = meta["bands"] + SENTINEL_L2_PRODUCTS["60m"]
 
     return meta
 
 
+def _l2_prefixed_band(band):
+    """Return L2A prefixed bands name."""
+    if band in SENTINEL_L2_BANDS["60m"]:
+        for res, bands in SENTINEL_L2_BANDS.items():
```
DanSchoppe: This strategy relies on the order of key/value pairs in the dictionary to find the highest resolution for each band. Based on some quick research, it appears Python versions prior to 3.7 did not track dictionary insertion order, and therefore the order in which elements are returned from iterators like `.items()` cannot be relied upon. Unless I'm reading the docs wrong, which is entirely possible 😂

vincentsarago: Oh no I think you are right ;-)
I'll try to come up with tests and use https://docs.python.org/2/library/collections.html#collections.OrderedDict

DanSchoppe:
😍 Looks good! Thanks, @vincentsarago for putting this together!!

rio_tiler/sentinel2.py
```diff
@@ -14,64 +16,81 @@
...
+SENTINEL_L2_BANDS = {
+    "10m": ["02", "03", "04", "08"],
+    "20m": ["02", "03", "04", "05", "06", "07", "08", "11", "12", "8A"],
+    "60m": ["01", "02", "03", "04", "05", "06", "07", "08", "09", "11", "12", "8A"],
+}
+
+SENTINEL_L2_PRODUCTS = {
+    "10m": ["AOT", "WVP"],
+    "20m": ["AOT", "SCL", "WVP"],
+    "60m": ["AOT", "SCL", "WVP"],
```
DanSchoppe: @vincentsarago I'm assuming the reason `TCI` isn't exposed here is because it's a 3-band dataset, unlike all the other single-band options? I suppose that would throw a wrench into the current rio-tiler strategy.

It's a bit unfortunate that TCI isn't trivial to support. I saw in your recent [COG Talk 4ter - Distributed processes](https://medium.com/devseed/cog-talk-4ter-distributed-processes-8ee280f71080) that you're specifying rio-color params to get bands 04,03,02 to look good. I use a similar strategy but my results are usually significantly less pleasing than these TCI images 😄.

I can imagine someday adding an additional arg to `sentinel2.tile()` that could allow the user to request the TCI imagery. Or even splitting TCI into three individual bands that could be specified in the bands arg `bands=("TCI1", "TCI2", "TCI3")`

vincentsarago: yes this is intentional because you can easily read the TCI using `rio_tiler.main.tile` function. 

```
from rio_tiler.sentinel2 import _sentinel_parse_scene_id
from rio_tiler.main import tile

tci_path = os.path.join(
        scene_params["aws_bucket"],
        scene_params["aws_prefix"],
        "R10m",
        "TCI.jp2"
)
t, m = tile(tci_path, x, y, z)
```

Basically, if we were adding multiple bands file support (e.g TCI) in addition to single band it will be a nightmare to handle.

DanSchoppe: > if we were adding multiple bands file support (e.g TCI) in addition to single band it will be a nightmare to handle

Yes, totally, I understand completely.

Thanks for the suggestion. That's a great strategy.

PR #155, opened 2020-03-13
Terracotta default transform

vincentsarago:
@dionhaefner, finally bringing your calculate_default_tranform 🙏

@rowanwins @DanSchoppe I've set back the padding to 0 as default value because I still don't see obvious change on the border of the image when we use padding or not. What I have been able to see it that sometime the image will appear a bit more blurry (on the whole image) when using padding.


dionhaefner:
:+1: I suggest you try this on an image from northern Greenland. The difference in blurriness should be striking.

vincentsarago:
thanks @dionhaefner this is definitely way faster (benchmark confirmed this too).

rowanwins:
No worries @vincentsarago , I think that's sensible

PR #182, opened 2020-05-18
allow reading part of data at hight resolution and add area/point function in io.cogeo

vincentsarago:
This PR does:
- allow reading high resolution part of a raster (by making height, width args optional)
- deprecated `out_window` in favor of `window` in `rio_tiler.reader._read`
- add `point` and `area` function in `rio_tiler.io.cogeo`
- add `max_size` option in `rio_tiler.reader.part` to fix a maximum output size when `height` and `width` are not set

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -18,7 +18,7 @@ from rasterio.crs import CRS
 from rasterio.vrt import WarpedVRT
 from rasterio.warp import transform_bounds, transform
 from rasterio.enums import Resampling, ColorInterp
-from rasterio.windows import Window
+from rasterio import windows
 
 from rio_tiler import constants
 from rio_tiler.utils import (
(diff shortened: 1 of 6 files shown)


tests/test_io_cogeo.py
```diff
@@ -31,6 +32,7 @@ def test_spatial_info_valid():
     assert meta.get("minzoom")
     assert meta.get("maxzoom")
     assert meta.get("center")
+    print(meta.get("center"))
```
vincentsarago, line 35: 🤦

rio_tiler/reader.py
```diff
@@ -203,18 +220,30 @@ def part(
             )
 
     vrt_transform, vrt_width, vrt_height = get_vrt_transform(
-        src_dst, bounds, width, height, dst_crs=dst_crs
+        src_dst, bounds, height, width, dst_crs=dst_crs
```
vincentsarago, line 223: BUG

PR #186, opened 2020-05-27
refactor colormaps

vincentsarago:
This PR does:
- move colormaps files to `cmap_data/` (instead of `cmap/`) 
- add `rio_tiler.colormap.cmap` object to hold colormap file path and custom dict
- allow registration of colormaps 

```python
from rio_tiler.colormap import cmap

cm = {0: [], 1: [], 2:[]}
cmap.register('custom', cm)

cmap.get('custom') 
> {0: [], 1: [], 2: []}

- add depreciation notice for `rio_tiler.colormap.get_colormap`

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -62,14 +62,35 @@ print(mask.shape)
...
 from rio_tiler.utils import render
 
 buffer = render(tile, mask=mask) # this returns a buffer (PNG by default)
 ```
 
+Rescale non-byte data and apply colormap
+
+```python
+from rio_tiler.colormap import cmap
+from rio_tiler.utis import linear_rescale
```
kylebarron, line 77: utils

README.md
```diff
@@ -62,14 +62,35 @@ print(mask.shape)
...
+
+for bdx in range(tile.shape[0]):
```
kylebarron, line 79: I think you should add a comment line here to make it clear what this is doing, e.g.
```
# Rescale each band linearly to the unsigned byte range of 0-255
```
Also do you have to provide `in_range` when you're rescaling? If not passed does it automatically find the `min/max` of the band?

README.md
```diff
@@ -74,17 +74,12 @@ Rescale non-byte data and apply colormap
...
-        ),
-        0,
-    )
-tile = tile.astype(numpy.uint8)
+from rio_tiler.utils import linear_rescale
+
+# Rescale the tile array only where mask is valid and cast it to byte
+tile = numpy.where(
+  mask, linear_rescale(tile, in_range=(0, 1000), out_range=[0, 255]), 0
+).astype(numpy.uint8)
```
kylebarron, line 82: 👍 I think that's a clearer example for the front page docs

PR #189, opened 2020-05-29
refactor `rio_tiler.utils.tile_exits` to allow raster bounds latitude == -90,90

vincentsarago:
This pr refactor `rio_tiler.utils.tile_exits` to make sure rio-tiler doesn't fails when a raster as a bounds == -90 or 90 latitude 

cc @abarciauskas-bgse


kylebarron:
Is this `contains` or `intersects`?

vincentsarago:
@kylebarron good question

kylebarron:
From a cursory look it looked like contains

CHANGES.txt
```diff
@@ -1,3 +1,7 @@
+2.0a11 (2020-05-29)
+------------------
+- refactor `rio_tiler.utils.tile_exits` to allow raster bounds latitude == -90,90
```
abarciauskas-bgse, line 3: Minor but I think this should be `tile_exists`

kylebarron: I was pretty confused about where the tiles were going!

vincentsarago:
thanks @kylebarron and @abarciauskas-bgse 
I added more tests to confirm the code does intersects and contains 🎉

PR #204, opened 2020-07-27
add mosaic submodule

vincentsarago:
This PR takes code from rio-tiler-mosaic to add it as a submodule


README.md
```diff
@@ -370,9 +370,35 @@ print(tile.shape)
 > (1, 256, 256)
 ```
 
-#### Working with multiple assets
+## Working with multiple assets
 
-`rio_tiler.io.cogeo` submodule has `multi_*` functions (tile, part, preview, point, metadata, info, stats) allowing to fetch and merge info 
+#### Mosaic
+
+Starting in rio-tiler 2.0, we've transfered the [rio-tiler-mosaic](https://github.com/cogeotiff/rio-tiler-mosaic) plugin to be a rio-tiler submodule.
```
kylebarron, line 377: `transferred`. I'd also suggest saying "a rio-tiler module", rather than "submodule" because the latter makes me think of a Git submodule. I was actually confused before that the code was all copied in, because from the title I was expecting a git submodule

README.md
```diff
@@ -370,9 +370,35 @@ print(tile.shape)
...
+
+```python
+from rio_tiler.io import COGReader
+from rio_tiler.mosaic import mosaic_reader
+from rio_tiler.mosaic.methods import defaults
+
+
+def tiler(src_path: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
+    with COGReader(src_path) as cog:
```
kylebarron, line 386: It would be nice if you could pass `COGReader` directly to `mosaic_reader` instead of having to make a wrapper function.

README.md
```diff
@@ -370,9 +370,35 @@ print(tile.shape)
...
+        return cog.tile(*args, **kwargs)
+
+assets = ["mytif1.tif", "mytif2.tif", "mytif3.tif"]
+tile, mask = mosaic_reader(assets, tiler, 1, 1, 1)
+```
+
+You can also
```
kylebarron, line 393: also what??? The suspense!

vincentsarago: removed

README.md
vincentsarago: Yes BUT, then you assume things about `reader` function in https://github.com/cogeotiff/rio-tiler/blob/ce56ecfce934db3078944a7477e2f885a2d09474/rio_tiler/mosaic/reader.py#L46 

Don't forget the new `mosaic_reader` can be used for more then `tile`

vincentsarago: also check the last example in https://github.com/cogeotiff/rio-tiler/blob/master/Notebook/Using-rio-tiler-mosaic.ipynb

kylebarron: Yeah I kinda envision a `MosaicReader` class similar to the `COGReader`, like

```py
with MosaicReader(assets, COGReader, x, y, z) as mosaic:
	tile = mosaic.tile()
	part = mosaic.part()
	point = mosaic.point()
```

or in your last example in that notebook, you use two different tiling functions, so you could have:
```py
with MosaicReader(assets, x, y, z) as mosaic:
	tile, mask = mosaic.tile(STACReader, assets="visual")
	ndvi, _ = mosaic.tile(STACReader, expression="(B08-B04)/(B08+B04)")
	return numpy.concatenate((tile, ndvi)), mask
```

That seems a little cleaner and less complex than having a wrapper function. And the `MosaicReader` would use the `.tile` or `.point` etc function provided by the class in the first argument.

vincentsarago: looks good to me! we could just ask for the `reader` function to be either a COGReader or STACReader class

kylebarron: Yeah I suggested a context manager because I know you're obsessed with them 😉

Issue #203, opened 2020-07-23
should spatial_info be a @property ? 


vincentsarago:
in 
https://github.com/cogeotiff/rio-tiler/blob/94e55538c5ff8e8e70ffd38b8011f471ec870d1c/rio_tiler/io/cogeo.py#L236

vincentsarago:
changed in #206

PR #206, opened 2020-07-28
use abc base class for COGReader and STACReader

### LINKED ISSUE #205: Adding MosaicReader ContextManager
This class should take a COGReader/STACReader class as input so the user doesn't have to create the wrapper on their side.

e.g: 
```python
@dataclass
class MosaicReader:
    """Mosaic of COGs"""

    assets: Sequence[str]
    reader: Union[COGReader, STACReader]

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        pass

    def tile(self, x: int, y: int, z: int, **kwargs: Any):
        """Read tile"""
        def _reader(asset, x: int, y: int, z: int, **kwargs: Any):
            with self.reader(asset) as cog:
                return cog.tile(x, y, z, **kwargs)

        return mosaic_reader(self.assets, _reader, x, y, z, **kwargs)

    def part(self, bbox: Tuple[float, float, float, float], **kwargs: Any):
        """Read tile"""
        def _reader(asset, bbox: Tuple[float, float, float, float], **kwargs: Any):
            with self.reader(asset) as cog:
                return cog.tile(bbox, **kwargs)

        return mosaic_reader(self.assets, _reader, bbox, **kwargs)

    def preview(self, **kwargs: Any):
        """Read tile"""
        def _reader(**kwargs):
            with self.reader(asset) as cog:
                return cog.preview(**kwargs)

        return mosaic_reader(self.assets, _reader, **kwargs)
```

vincentsarago:
closes #205 #203 

This PR does: 
- add rio_tiler.io.BaseReader ABC class for COGReader and STACReader to inherit from
- make BaseReader.spatial_info a property (see #203)
- ~add rio_tiler.mosaic.MosaicReader class to avoid the user to create wrapper around their tiler (see #205)~

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -1,0 +1,101 @@
+"""rio_tiler.io.base: ABC class for rio-tiler readers."""
+
+import abc
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Sequence, Tuple
+
+import numpy
+
+
+# ref: https://github.com/python/mypy/issues/5374
+@dataclass  # type: ignore
+class BaseReader(metaclass=abc.ABCMeta):
+    """Rio-tiler.io BaseReader."""
+
+    bounds: Sequence[float] = field(init=False)
+
+    @abc.abstractmethod
+    def __enter__(self):
+        """Support using with Context Managers."""
+        ...
+
+    @abc.abstractmethod
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Support using with Context Managers."""
+        ...
+
+    @property
+    def minzoom(self) -> int:
+        """Dataset Min zoom."""
+        ...
+
+    @property
+    def maxzoom(self) -> int:
+        """Dataset Max zoom."""
+        ...
+
+    @property
+    def center(self) -> Tuple[float, float, int]:
+        """Dataset center + minzoom."""
+        return (
+            (self.bounds[0] + self.bounds[2]) / 2,
+            (self.bounds[1] + self.bounds[3]) / 2,
+            self.minzoom,
+        )
+
+    @property
+    def info(self) -> Dict:
+        """Return Dataset's info."""
+        ...
+
+    @property
+    def spatial_info(self) -> Dict:
+        """Return Dataset's spatial info."""
+        return {
+            "bounds": self.bounds,
+            "center": self.center,
+            "minzoom": self.minzoom,
+            "maxzoom": self.maxzoom,
+        }
+
+    @abc.abstractmethod
+    def stats(
+        self,
+        pmin: float = 2.0,
+        pmax: float = 98.0,
+        hist_options: Dict = {},
+        **kwargs: Any,
+    ) -> Dict:
+        """Return Dataset's statistics."""
+        ...
+
+    @abc.abstractmethod
+    def metadata(self, pmin: float = 2.0, pmax: float = 98.0, **kwargs: Any,) -> Dict:
+        """Return Dataset's statistics and info."""
+        info = self.info.copy()
+        info["statistics"] = self.stats(pmin, pmax, **kwargs)
+        return info
+
+    @abc.abstractmethod
+    def tile(
+        self, tile_x: int, tile_y: int, tile_z: int, **kwargs: Any
+    ) -> Tuple[numpy.ndarray, numpy.ndarray]:
+        """Read a Map tile from the Dataset."""
+        ...
+
+    @abc.abstractmethod
+    def part(
+        self, bbox: Tuple[float, float, float, float], **kwargs: Any
+    ) -> Tuple[numpy.ndarray, numpy.ndarray]:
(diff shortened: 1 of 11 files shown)


rio_tiler/io/base.py
```diff
@@ -0,0 +1,101 @@
...
+
+import numpy
+
+
+# ref: https://github.com/python/mypy/issues/5374
+@dataclass  # type: ignore
+class BaseReader(metaclass=abc.ABCMeta):
+    """Rio-tiler.io BaseReader."""
+
+    bounds: Sequence[float] = field(init=False)
```
vincentsarago, line 15: bounds will be set in `__enter__`, I'm not sure if there is a best way to tell the class

rio_tiler/io/base.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @property
+    def spatial_info(self) -> Dict:
+        """Return Dataset's spatial info."""
+        return {
+            "bounds": self.bounds,
+            "center": self.center,
+            "minzoom": self.minzoom,
+            "maxzoom": self.maxzoom,
+        }
```
vincentsarago, line 59: Breaking change

rio_tiler/io/stac.py
```diff
@@ -337,22 +329,29 @@ def point(
 
     def stats(
         self,
-        assets: Union[Sequence[str], str],
         pmin: float = 2.0,
         pmax: float = 98.0,
+        hist_options: Dict = {},
+        assets: Union[Sequence[str], str] = None,
```
vincentsarago, line 335: This change was made to accommodate the ABC class and avoid changing the `args`.

Note: Breacking change

rio_tiler/mosaic/reader.py
```diff
@@ -41,6 +43,76 @@ def _filter_tasks(tasks: TaskType):
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Support using with Context Managers."""
+        pass
+
+    def tile(self, x: int, y: int, z: int, **kwargs: Any):
+        """Get Tile from multiple observation."""
+
+        def _reader(asset, *args: int, **kwargs: Any):
+            with self.reader(asset) as src_dst:  # type: ignore
```
vincentsarago, line 92: I don't fully understand why mypy is not happy here.

rio_tiler/mosaic/reader.py
```diff
@@ -41,6 +43,76 @@ def _filter_tasks(tasks: TaskType):
...
+        Return map tile from multiple observation.
+    part((0,10,0,10), **kwargs)
+        Return array for a specific bbox from multiple observation.
+    preview(**kwargs)
+        Return a preview from multiple observation.
+
+    """
+
+    assets: Sequence[str]
+    reader: BaseReader
```
vincentsarago, line 78: not sure if we want to make COGReader the default here?

geospatial-jeff: I think `BaseReader` is good here

rio_tiler/io/stac.py
```diff
@@ -337,22 +329,29 @@ def point(
...
         **kwargs: Any,
     ) -> Dict:
         """Return array statistics from COGs."""
+        if not assets:
+            raise InvalidBandName("Missing 'assets'")
```
geospatial-jeff, line 340: Should `InvalidBandName` be renamed to `InvalidAsset` or similar?  Assets are not always bands.

rio_tiler/mosaic/reader.py
```diff
@@ -41,6 +43,76 @@ def _filter_tasks(tasks: TaskType):
             pass
 
 
+@dataclass
+class MosaicReader:
```
geospatial-jeff, line 47: I don't think this is possible right now, there is too much collision between the base classes, but it would be nice for `MosaicReader` to extend `cogeo_mosaic.backends.BaseBackend`.  This way the user can naively pass a mosaicjson into `MosaicReader` and read a tile without having to worry about asset selection.

rio_tiler/io/stac.py
vincentsarago: Yep we should have `InvalidAssetName` and `MissingAssetName` will add that in another PR

rio_tiler/mosaic/reader.py
vincentsarago: > Maybe this is better off in another lib, I can see the argument where rio-tiler should only be concerned with reading tiles from image(s).

Yes I will tend to agree with that.

vincentsarago:
following https://github.com/cogeotiff/rio-tiler/pull/206#discussion_r462658831 we've decided to remove the MosaicReader from rio-tiler and add new features directly to cogeo-mosaic. While the branch is named `MosaicReader` this PR just do:

- add rio_tiler.io.BaseReader ABC class for COGReader and STACReader to inherit from
- make BaseReader.spatial_info a property (see #203)

PR #210, opened 2020-07-30
update STACReader exceptions

### LINKED ISSUE #208: rename InvalidBandName to `InvalidAssetName` and `MissingAssetName` for STAC
we should have `InvalidAssetName` and `MissingAssetName` instead of InvalidBandName in STACReader

_Originally posted by @vincentsarago in https://github.com/cogeotiff/rio-tiler/pull/206#discussion_r462669202_

vincentsarago:
closes #208

### DIFF
--- a/rio_tiler/io/stac.py
+++ b/rio_tiler/io/stac.py
@@ -183,9 +183,9 @@ class STACReader(BaseReader):
 
     def _get_href(self, assets: Sequence[str]) -> Sequence[str]:
         """Validate asset names and return asset's url."""
-        for asset in assets:
-            if asset not in self.assets:
-                raise InvalidAssetName(f"{asset} is not a valid asset name.")
+        invalid_assets = set(assets) - set(self.assets)
+        if invalid_assets:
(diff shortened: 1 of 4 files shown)


rio_tiler/io/stac.py
```diff
@@ -185,7 +185,7 @@ def _get_href(self, assets: Sequence[str]) -> Sequence[str]:
         """Validate asset names and return asset's url."""
         for asset in assets:
             if asset not in self.assets:
-                raise InvalidBandName(f"{asset} is not a valid asset name.")
+                raise InvalidAssetName(f"{asset} is not a valid asset name.")
```
geospatial-jeff, line 188: I think we can prevent looping through the assets twice.  The result set will also contain all assets which are invalid.

```python
def _get_href(self, assets: Sequence[str]) -> Sequence[str]:
    invalid_assets = set(assets) - set(self.assets)
    if invalid_assets:
        raise InvalidAssetName(f"{invalid_assets} is not a valid asset name")
    return [self.item["assets"][asset]["href"] for asset in assets]
```

PR #215, opened 2020-07-31
add "used_assets" to the mosaic reader output

vincentsarago:
ref: https://github.com/developmentseed/titiler/issues/51

**This is a breaking change.**

With this PR the mosaic_reader now returns `((tile, mask), used_assets)`


kylebarron:
> now returns `((tile, mask), used_assets)`

Just to be clear this means you can't destructure with
```py
tile, mask, asset_names = cog.tile(...)
```
because there's two levels of nesting. You have to do 
```py
(tile, mask), asset_names = cog.tile(...)
```

vincentsarago:
> because there's two levels of nesting.

@kylebarron absolutely, is this a blocker? 

FYI, this is because of https://github.com/cogeotiff/rio-tiler/blob/a0d9ead02eaa8c586629fbae2dc61d4e04788592/rio_tiler/mosaic/reader.py#L130

we return a tuple (from pixel_selection.data) and a variable (assets_used)

PR #228, opened 2020-08-13
add `nodata`, `unscale` and `vrt_options` in Reader __init__

### LINKED ISSUE #226: Add `External` to indicate the Nodata was passed in **kwargs
https://github.com/cogeotiff/rio-tiler/blob/065e9d6079e651a1488709a0df6db849374f39c6/rio_tiler/reader.py#L553-L560

vincentsarago:
closes #226

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -232,7 +232,7 @@ class COGReader(BaseReader):
             Dictionary with bands statistics.
 
         """
-        kwargs = {**kwargs, **self._kwargs}
+        kwargs = {**self._kwargs, **kwargs}
 
         hist_options = hist_options or {}
 
@@ -287,7 +287,7 @@ class COGReader(BaseReader):
         mask: numpy array
 
         """
-        kwargs = {**kwargs, **self._kwargs}
+        kwargs = {**self._kwargs, **kwargs}
 
         if isinstance(indexes, int):
             indexes = (indexes,)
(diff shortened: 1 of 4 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -135,6 +138,19 @@ def maxzoom(self) -> int:
...
+            kwargs["nodata"] = nodata
+
+        vrt_options = kwargs.get("vrt_options", self.vrt_options)
+        if vrt_options is not None:
+            kwargs["vrt_options"] = vrt_options
+
+        unscale = kwargs.get("unscale", self.unscale)
+        if unscale is not None:
+            kwargs["unscale"] = unscale
+
```
vincentsarago, line 153: Please, if there is a better way, feel free to edit this.

rio_tiler/reader.py
```diff
@@ -323,6 +323,7 @@ def point(
     nodata: Optional[Union[float, int, str]] = None,
     unscale: bool = False,
     masked: bool = True,
+    vrt_options: Optional[Dict] = None,
```
vincentsarago, line 326: vrt_options was missing in `point`

rio_tiler/io/cogeo.py
geospatial-jeff: I think we can use `__post_init__` (https://docs.python.org/3/library/dataclasses.html#post-init-processing)

PR #231, opened 2020-08-18
Simplify STAC's get_assets if/else

kylebarron:


kylebarron:
Ideally this PR hoped to simplify the branching for `get_assets`. This is a hard problem because there are _two_ sets of include/excludes: one for the filenames and another for the file _types_. It's hard to accurately do two sets of include/excludes. Which takes preference? When do you use AND, when do you use OR?

I'm not currently using STAC. I would guess that this topic will come up again, at which point coming back to this PR could be a good idea.

PR #243, opened 2020-08-26
Try to fix mkdocs deploy

kylebarron:


vincentsarago:
So I fixed in in https://github.com/cogeotiff/rio-tiler-pds. 
The problem is the actions is trying to install rio-tiler but we need numpy to be installed first

```yml
jobs:
  build:
    name: Deploy docs
    runs-on: ubuntu-latest
    steps:
      - name: Checkout master
        uses: actions/checkout@v2

      - name: Set up Python 3.8
        uses: actions/setup-python@v2
        with:
          python-version: 3.8

      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip
          python -m pip install numpy
          python -m pip install -e .["docs"]

      - name: Deploy docs
        run: mkdocs gh-deploy --force
```

and in setup.py I added 
```
"docs": ["mkdocs", "mkdocs-material"],
```

PR #247, opened 2020-08-29
Remove deprecated functions

kylebarron:
Regarding removing `rio_tiler.mosaic.mosaic_tiler`: I feel that if you’re doing a major version bump anyways, there’s no reason to re-export a function with two names.

In any case, `mosaic_tiler` isn't currently backwards compatible because it returns the output of `mosaic_reader` directly, but `mosaic_reader` was updated to return the `asset_ids` of the assets that were used. If you decide to keep `mosaic_tiler` for backwards compatibility with `rio_tiler_mosaic`, it would be good to return only the `tile, mask`.

### DIFF
--- a/rio_tiler/mosaic/reader.py
+++ b/rio_tiler/mosaic/reader.py
@@ -87,28 +87,3 @@ def mosaic_reader(
                 return pixel_selection.data, assets_used
 
     return pixel_selection.data, assets_used
-
-
-def mosaic_tiler(
-    assets: Sequence[str],
-    tile_x: int,
-    tile_y: int,
-    tile_z: int,
-    reader: Callable,
-    pixel_selection: Optional[MosaicMethodBase] = None,
-    chunk_size: Optional[int] = None,
-    threads: int = MAX_THREADS,
-    **kwargs,
-) -> Tuple[Tuple[numpy.ndarray, numpy.ndarray], Sequence[str]]:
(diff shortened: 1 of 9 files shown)


rio_tiler/mosaic/reader.py
```diff
@@ -87,28 +87,3 @@ def reader(asset: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
...
-        assets,
-        reader,
-        tile_x,
-        tile_y,
-        tile_z,
-        pixel_selection=pixel_selection,
-        chunk_size=chunk_size,
-        threads=threads,
-        **kwargs,
-    )
```
vincentsarago, line 114: 👍

rio_tiler/reader.py
```diff
@@ -247,13 +246,6 @@ def part(
             height=orig_vrt_height,
         )
 
-    if warp_vrt_option:
-        warnings.warn(
-            "warp_vrt_option will be removed in 2.0, use vrt_options",
-            DeprecationWarning,
-        )
-        vrt_options = vrt_options or warp_vrt_option
-
```
vincentsarago, line 256: I think you didn't remove the options in the function definition!

kylebarron: oh yeah you're right

vincentsarago:
@kylebarron i'll let you fix the merge conflict and merge

PR #250, opened 2020-08-29
Allow class (not just instance) to be passed to pixel_selection

kylebarron:
This PR allows a _class_ subclassed from `MosaicMethodBase` to be passed to `pixel_selection`, not just an instance of that class.

Currently the `mosaic_reader` and `mosaic_tiler` classes expect `pixel_selection` to be an _instance_ of `MosaicMethodBase`. This leads to some ugly code (IMO):

```py
tile, mask = mosaic_reader(
    assets,
    reader,
    x,
    y,
    z,
    pixel_selection=FirstMethod()
)
```

If you forget to instantiate the class, and instead pass `pixel_selection=FirstMethod`, you see the error:
```
Mosaic filling algorithm should be an instance of'rio_tiler.mosaic.methods.base.MosaicMethodBase'
```
which isn't clear to users.

The docstring also currently says:
```
pixel_selection: MosaicMethod, optional
    Instance of MosaicMethodBase class.
    default: "rio_tiler.mosaic.methods.defaults.FirstMethod".
```

and the default leads one to think it doesn't need to be called.


rio_tiler/mosaic/reader.py
```diff
@@ -59,6 +59,9 @@ def reader(asset: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
         Return (tile, mask) data and list of assets used.
 
     """
+    if issubclass(pixel_selection, MosaicMethodBase):
```
vincentsarago, line 62: If think it will raise an error if pixel_selection is not a class 
```
import inspect
if inspect.isclass(pixel_selection) and issubclass(pixel_selection, MosaicMethodBase):
    pixel_selection = pixel_selection()
```

kylebarron:
I think now mypy is unhappy. How can you say you'd accept either the class or an instance of the class?

rio_tiler/mosaic/reader.py
```diff
@@ -16,7 +17,7 @@ def mosaic_reader(
     assets: Sequence[str],
     reader: Callable,
     *args: Any,
-    pixel_selection: Optional[MosaicMethodBase] = None,
+    pixel_selection: Union[Type[MosaicMethodBase], MosaicMethodBase, None] = None,
```
vincentsarago, line 20: Can this be Optional[Union[Type[MosaicMethodBase], MosaicMethodBase]] ?

rio_tiler/mosaic/reader.py
```diff
@@ -59,12 +60,18 @@ def reader(asset: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
         Return (tile, mask) data and list of assets used.
 
     """
+    if isclass(pixel_selection):
+        pixel_selection = cast(Type[MosaicMethodBase], pixel_selection)
```
vincentsarago, line 64: Not sure to understand why we need this!

kylebarron: Mypy errors without this because the input can be either a class or an instance and it fails on an instance. Mypy isn't smart enough to understand the `isclass and`

kylebarron: So it's telling mypy that within the if block, it's a class not an instance

rio_tiler/mosaic/reader.py
kylebarron: Sure. Sometimes the nesting can get verbose

rio_tiler/mosaic/reader.py
vincentsarago: what do you think about adding `# type: ignore` instead of this new if block ?

kylebarron: It's roughly equivalent but the `cast` is more explicit; it's still checking instead of just ignoring

vincentsarago: Ok, at this point I think it might be better to pass a class instead of an object 🤷‍♂️ (this will be a breaking change but it might be cleaner)

PR #260, opened 2020-09-09
add resampling_method in COGReader global options

vincentsarago:
This PR does:

- remove useless `hist_options` in `BaseReader.stats` method
- add `resampling_method` in COGReader global options
- add `resampling_method` in COGReader.point for compatibility with `_read`
- re-order keyword argument in `reader._read` and `reader.point`

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -34,10 +34,10 @@ def _read(
     width: Optional[int] = None,
     indexes: Optional[Union[Sequence[int], int]] = None,
     window: Optional[windows.Window] = None,
-    nodata: Optional[Union[float, int, str]] = None,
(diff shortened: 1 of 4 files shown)


rio_tiler/reader.py
```diff
@@ -358,7 +359,10 @@ def point(
 
     indexes = indexes if indexes is not None else src_dst.indexes
 
-    vrt_params: Dict[str, Any] = {"add_alpha": True}
+    vrt_params: Dict[str, Any] = {
+        "add_alpha": True,
+        "resampling": Resampling[resampling_method],
```
vincentsarago, line 364: Note: this is not really needed because resampling should not have any effect on the `sample` command because it fetch directly the raw block. I added it so resampling_method is used somewhere in the function!

PR #262, opened 2020-09-12
add post_process callback to _read and point

### LINKED ISSUE #181: Add `postprocess` callback
Sometimes we want to apply post process on a tile (or data), like apply a custom rescaling.

We can easily achieve this by just wrapping the `tile` function: 

```python
from rio_tiler import reader

def my_tile(path, x, y, z):
    tile, master = reader.tile(path, x, y, z)

    return tile / 10000.0, mask
```

But when if you wanted to get the `metadata` of the same raster with your custom `process` added, you'll have to re-write the whole `rio_tiler.reader.metadata` function.

To overcome this we could *just* add a `postprocess: Callable` option in `rio_tiler.reader._read` function.

```python
def postprocess(arr: numpy.ndarray, mask: numpy.ndarray) -> Tuple[numpy.ndarray, numpy.ndarray]:
   
   arr = arr // 10000.0

   return arr, mask
```

Because `metadada`, `preview`, `part` and `tile` all use `_read` they will forward the `postprocess` function to `_read`.

```python
from rio_tiler import reader

with rasterio.open("my.tif") as src_dst:
     metadata = reader.metadata(src_dst, postprocess=postprocess)

```

Changes in `rio_tiler.reader._read`
```python
def _read(
    src_dst: Union[DatasetReader, DatasetWriter, WarpedVRT],
    height: int,
    width: int,
    indexes: Optional[Union[Sequence[int], int]] = None,
    out_window: Optional[Window] = None,
    nodata: Optional[Union[float, int, str]] = None,
    resampling_method: Resampling = "bilinear",
    force_binary_mask: bool = True,
    unscale: bool = False,
    vrt_options: Dict = {},
    postprocess: Optional[Callable] = None,
) -> Tuple[numpy.ndarray, numpy.ndarray]:

    if postprocess:
         return postprocess(data, mask)

    return data, mask
```

Question:
- How do we enfore `postprocess` callable to accept `data` and `mask` as minimum input ?
- How do we enfore `postprocess` callable to return only `data` and `mask`

vincentsarago:
closes #181 

This PR does:
- add `post_process` callback to `reader._read` and `reader.point` to apply specific operation on output array. This is mostly to be able to apply ops on arrays for `metadata` and `stats`. 
- refactor `reader.point` to accept the post_process option.

@kylebarron I'm happy to change the name of the option if you have a better idea. 

I'll also try to add an example using Landsat8 rio-tiler-pds reader to show how this can be useful.

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -2,7 +2,7 @@
 
 import warnings
 from concurrent import futures
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import attr
 import mercantile
@@ -95,8 +95,11 @@ class COGReader(BaseReader):
     # Define global options to be forwarded to functions reading the data (e.g rio_tiler.reader._read)
     nodata: Optional[Union[float, int, str]] = attr.ib(default=None)
     unscale: Optional[bool] = attr.ib(default=None)
-    vrt_options: Optional[Dict] = attr.ib(default=None)
     resampling_method: Optional[Resampling] = attr.ib(default=None)
+    vrt_options: Optional[Dict] = attr.ib(default=None)
+    post_process: Optional[
+        Callable[[numpy.ndarray, numpy.ndarray], Tuple[numpy.ndarray, numpy.ndarray]]
+    ] = attr.ib(default=None)
 
     # We use _kwargs to store values of nodata, unscale, vrt_options and resampling_method.
     # _kwargs is used avoid having to set those values on each method call.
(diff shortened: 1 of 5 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -95,8 +95,11 @@ class COGReader(BaseReader):
     # Define global options to be forwarded to functions reading the data (e.g rio_tiler.reader._read)
     nodata: Optional[Union[float, int, str]] = attr.ib(default=None)
     unscale: Optional[bool] = attr.ib(default=None)
-    vrt_options: Optional[Dict] = attr.ib(default=None)
     resampling_method: Optional[Resampling] = attr.ib(default=None)
+    vrt_options: Optional[Dict] = attr.ib(default=None)
+    post_process: Optional[
+        Callable[[numpy.ndarray, numpy.ndarray], Tuple[numpy.ndarray, numpy.ndarray]]
```
kylebarron, line 101: Is
```py
Callable[[numpy.ndarray, numpy.ndarray], Tuple[numpy.ndarray, numpy.ndarray]]
```

fine instead of
```py
Callable[Tuple[numpy.ndarray, numpy.ndarray], Tuple[numpy.ndarray, numpy.ndarray]]
```
?

kylebarron (approved):

rio_tiler/io/cogeo.py
vincentsarago: @kylebarron if we do that, then the postprocess callable has be be 
```python
data, mask = post_process((data, mask))
```
instead of 
```python
data, mask = post_process(data, mask)
```

I'm not a big fan of passing a tuple as input but if you think it makes more sense or is more user friendly 👌

kylebarron: Oh I see, no I agree I was just confused on the typings

PR #265, opened 2020-09-18
asyncio support

### LINKED ISSUE #264: Add AsyncBaseReader, AsyncCogReader
Async implementations of rio tiler readers using aiocogeo (ref https://github.com/geospatial-jeff/aiocogeo/issues/71).

geospatial-jeff:
Closes #264 @vincentsarago @kylebarron

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -117,6 +117,14 @@ class AsyncBaseReader(metaclass=abc.ABCMeta):
     minzoom: int = attr.ib(init=False)
     maxzoom: int = attr.ib(init=False)
 
+    async def __aenter__(self):
+        """Support using with Context Managers."""
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        """Support using with Context Managers."""
+        pass
+
     @property
     def center(self) -> Tuple[float, float, int]:
         """Dataset center + minzoom."""
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -116,13 +116,8 @@ class BaseReader(SpatialMixin, metaclass=abc.ABCMeta):
 
 
 @attr.s
-class AsyncBaseReader(metaclass=abc.ABCMeta):
+class AsyncBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
     """Rio-tiler.io AsyncBaseReader."""
-
-    tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
-    bounds: Tuple[float, float, float, float] = attr.ib(init=False)
-    minzoom: int = attr.ib(init=False)
-    maxzoom: int = attr.ib(init=False)
 
     async def __aenter__(self):
         """Support using with Context Managers."""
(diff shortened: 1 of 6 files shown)


rio_tiler/io/base.py
```diff
@@ -92,11 +82,40 @@ def point(self, lon: float, lat: float, **kwargs: Any) -> List:
         ...
 
 
+class SyncBaseReader(BaseReader, metaclass=abc.ABCMeta):
```
geospatial-jeff, line 85: Add two more classes to differentiate between sync/async

rio_tiler/io/base.py
```diff
@@ -92,6 +83,95 @@ def point(self, lon: float, lat: float, **kwargs: Any) -> List:
         ...
 
 
+@attr.s
+class AsyncBaseReader(metaclass=abc.ABCMeta):
```
geospatial-jeff, line 87: `abc` doesn't enforce the signature of the method, just that the method is present, so we could reuse the base class but mypy is not happy because the response type is different.

rio_tiler/io/base.py
```diff
@@ -21,16 +22,6 @@ class BaseReader(metaclass=abc.ABCMeta):
...
-    @abc.abstractmethod
-    def __enter__(self):
-        """Support using with Context Managers."""
-        ...
-
-    @abc.abstractmethod
-    def __exit__(self, exc_type, exc_value, traceback):
-        """Support using with Context Managers."""
-        ...
-
```
vincentsarago, line 33: 👍

vincentsarago (approved):
LGTM

geospatial-jeff:
I'd like to try this out in aiocogeo before we merge.  Started that work in a branch (https://github.com/geospatial-jeff/aiocogeo/tree/rio-tiler) but there are a few things I need to add first to match rio-tiler like color interp and color table.

rio_tiler/io/base.py
```diff
@@ -108,6 +109,103 @@ def tile_exists(self, tile_z: int, tile_x: int, tile_y: int) -> bool:
...
+        ...
+
+    @abc.abstractmethod
+    async def point(
+        self, lon: float, lat: float, **kwargs: Any
+    ) -> Coroutine[Any, Any, List]:
+        """Read a value from a Dataset."""
+        ...
+
+    def tile_exists(self, tile_z: int, tile_x: int, tile_y: int) -> bool:
```
geospatial-jeff, line 198: This could go into a mixin if we wanted to dedup with `BaseReader`

vincentsarago: not sure to understand this @geospatial-jeff

rio_tiler/io/base.py
kylebarron: So we just need to make sure that this base class maintains parity with the `BaseReader`?

rio_tiler/io/base.py
kylebarron: You can have a `TileExistsMixin` that defines `tile_exists` (and maybe `center` and `spatial_info`?) since those are the same across the `BaseReader` and the `AsyncBaseReader`. Then you add the mixin to both classes.

If you want to make the mixin 👍 , but not a strong preference.

kylebarron (approved):

rio_tiler/io/base.py
geospatial-jeff: Just that this method is the same across the sync and async base classes.  We could only define this method in one place by writing a mixin which is shared across both base classes.  But repeating code isn't that big a deal so I'm fine leaving it as is as well!

rio_tiler/io/base.py
geospatial-jeff: Yeah, there might be a better way to enforce the same type across both classes besides duck typing but I can't think of one that would make mypy happy.

rio_tiler/io/base.py
vincentsarago: the Mixin is a great idea! I'll make a PR and then we can merge this

PR #270, opened 2020-09-29
GeoMedianMethod and MaxIndexMethod for pixel selection

wouellette:
I added `GeoMedianMethod` and `MaxIndexMethod` as pixel selection methods.
The former is based on Dale Roberts' `hdmedians` package, a geometric median implementation in cython.

For `GeoMedianMethod`, I didn't want to add cython dependencies to rio-tiler, so currently it's just a naive numpy array assignment which is rather slow, but left it like this to spark the discussion on how to best go forward on this one, as I think pixel selection using the geomedian is rather useful. I contacted the Geoscience Australia project, who seem to internally use an `hdstats` package that offers an interface to n-dimensional array geomedian compositing: [https://github.com/GeoscienceAustralia/digitalearthau/issues/260](https://github.com/GeoscienceAustralia/digitalearthau/issues/260 ), but it's currently not on PyPI.

For the `MaxIndexMethod`, the idea is to expand the behavior of a max NDVI composite (which is itself missing) to any possible band combination.

### DIFF
--- a/rio_tiler/mosaic/methods/defaults.py
+++ b/rio_tiler/mosaic/methods/defaults.py
@@ -98,8 +98,7 @@ class MedianMethod(MosaicMethodBase):
         """Return data and mask."""
         if self.tile:
             tile = numpy.ma.median(numpy.ma.stack(self.tile, axis=0), axis=0)
-            if self.enforce_data_type:
-                tile = tile.astype(self.tile[0].dtype)
+            tile = tile.astype(self.tile[0].dtype) if self.enforce_data_type else tile
             return tile.data, ~tile.mask[0] * 255
         else:
             return None, None
@@ -115,6 +114,7 @@ class StdevMethod(MosaicMethodBase):
     def __init__(self, enforce_data_type=True):
         """Overwrite base and init Stdev method."""
         super(StdevMethod, self).__init__()
+        self.enforce_data_type = enforce_data_type
         self.tile = []
 
     @property
@@ -122,6 +122,7 @@ class StdevMethod(MosaicMethodBase):
         """Return data and mask."""
         if self.tile:
             tile = numpy.ma.std(numpy.ma.stack(self.tile, axis=0), axis=0)
+            tile = tile.astype(self.tile[0].dtype) if self.enforce_data_type else tile
             return tile.data, ~tile.mask[0] * 255
         else:
             return None, None
@@ -134,22 +135,26 @@ class GeoMedianMethod(MosaicMethodBase):
 
     def __init__(self, enforce_data_type=True):
         super(GeoMedianMethod, self).__init__()
+        self.enforce_data_type = enforce_data_type
         self.tile = []
 
     @property
     def data(self):
         """Return data and mask."""
         if self.tile:
-            # Call Dale's function here
+            # Convert with rescaling factor to float32 for input to Dale's hdmedians
             stacked_tile = (numpy.ma.stack(self.tile, axis=1) / 10000).astype(numpy.float32)
             b, t, y, x = stacked_tile.shape
             tile = numpy.ma.zeros((b, y, x)).astype(numpy.float32)
 
+            # TODO: How to implement geomedian without introducing too many new dependencies?
             for xid in range(x):     # for each x-axis
                 for yid in range(y): # for each y-axis
                     tile[:, yid, xid] = hd.geomedian(stacked_tile[:, :, yid, xid])
 
-            return (tile * 10000).astype(numpy.uint16), (tile[0] != 0) * 255
+            # Convert back to uint16
+            tile = (tile * 10000).astype(self.tile[0].dtype) if self.enforce_data_type else (tile * 10000)
+            return tile.data, (tile[0] != 0) * 255
         else:
             return None, None
 
@@ -162,6 +167,7 @@ class MaxIndexMethod(MosaicMethodBase):
     def __init__(self, enforce_data_type=True, formula='(b2-b1)/(b2+b1)'):
         super(MaxIndexMethod, self).__init__()
         self.tile = []
+        self.enforce_data_type = enforce_data_type
         self.formula = formula
 
     @property
@@ -182,7 +188,9 @@ class MaxIndexMethod(MosaicMethodBase):
             argmax_index = numpy.ma.argmax(index, axis=0)
             ixgrid = numpy.ix_(numpy.arange(t), numpy.arange(y), numpy.arange(x))
             tile = tile[argmax_index, :, ixgrid[1], ixgrid[2]].squeeze(axis=0)
-            return numpy.moveaxis(tile.data, -1, 0), numpy.moveaxis(~tile.mask, -1, 0)[0] * 255
+            tile = numpy.moveaxis(tile, -1, 0).astype(self.tile[0].dtype) if self.enforce_data_type \
+                else numpy.moveaxis(tile, -1, 0)
+            return tile.data, ~tile.mask[0] * 255
         else:
             return None, None
 
--- a/rio_tiler/mosaic/methods/defaults.py
+++ b/rio_tiler/mosaic/methods/defaults.py
@@ -3,6 +3,7 @@
 import numpy
 
 from .base import MosaicMethodBase
+import hdmedians as hd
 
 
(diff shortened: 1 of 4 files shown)


rio_tiler/mosaic/methods/defaults.py
```diff
@@ -114,17 +114,86 @@ class StdevMethod(MosaicMethodBase):
...
+            tile = (tile * 10000).astype(self.tile[0].dtype) if self.enforce_data_type else (tile * 10000)
+            return tile.data, (tile[0] != 0) * 255
+        else:
+            return None, None
+
+    def feed(self, tile):
+        """Add data to tile."""
+        self.tile.append(tile)
+
+class MaxIndexMethod(MosaicMethodBase):
```
vincentsarago, line 165: I feel this method is a bit complex. I have used custom methods called `CustomLastBand...` 

The idea is to use `expression` like `B1,B2,B3,{NDVI}` which will create a 4 bands array and then within the mosaic method, we will exclude the last band from the array. 

```python
class CustomLastBandHigh(MosaicMethodBase):
    """Feed the mosaic tile using the last band as decision factor."""

    @property
    def data(self):
        """Return data and mask."""
        if self.tile is not None:
            return self.tile.data[:-1], ~self.tile.mask[0] * 255
        else:
            return None, None

    def feed(self, tile: numpy.ma.array):
        """Add data to tile."""
        if self.tile is None:
            self.tile = tile
            return

        pidex = (
            numpy.bitwise_and(tile.data[-1] > self.tile.data[-1], ~tile.mask)
            | self.tile.mask
        )

        mask = numpy.where(pidex, tile.mask, self.tile.mask)
        self.tile = numpy.ma.where(pidex, tile, self.tile)
        self.tile.mask = mask


class CustomLastBandLow(MosaicMethodBase):
    """Feed the mosaic tile using the last band as decision factor."""

    @property
    def data(self):
        """Return data and mask."""
        if self.tile is not None:
            return self.tile.data[:-1], ~self.tile.mask[0] * 255
        else:
            return None, None

    def feed(self, tile: numpy.ma.array):
        """Add data to tile."""
        if self.tile is None:
            self.tile = tile
            return

        pidex = (
            numpy.bitwise_and(tile.data[-1] < self.tile.data[-1], ~tile.mask)
            | self.tile.mask
        )

        mask = numpy.where(pidex, tile.mask, self.tile.mask)
        self.tile = numpy.ma.where(pidex, tile, self.tile)
        self.tile.mask = mask
```

rio_tiler/mosaic/methods/defaults.py
```diff
@@ -114,17 +114,86 @@ class StdevMethod(MosaicMethodBase):
...
+        if self.tile:
+            # Convert with rescaling factor to float32 for input to Dale's hdmedians
+            stacked_tile = (numpy.ma.stack(self.tile, axis=1) / 10000).astype(numpy.float32)
+            b, t, y, x = stacked_tile.shape
+            tile = numpy.ma.zeros((b, y, x)).astype(numpy.float32)
+
+            # TODO: How to implement geomedian without introducing too many new dependencies?
+            for xid in range(x):     # for each x-axis
+                for yid in range(y): # for each y-axis
+                    tile[:, yid, xid] = hd.geomedian(stacked_tile[:, :, yid, xid])
```
vincentsarago, line 153: Yeah that's won't be fast 😬 

I wonder if we could vectorize the calculation with numpy ufunc (cc @kylebarron)

kylebarron: I agree this won't be fast.

I think the main problem here is that `hd.geomedian` only supports a two-dimensional array as input. So it can't natively broadcast over a higher dimensional array.

Also how does this work with missing data? Does it automatically mask out portions of the array that are undefined?

wouellette: About the missing data, It returns `0s` currently, which you can then convert to `NaNs` _a posteriori_, or directly use the `hd.nangeomedian`, which handles missing data directly (should have used that one for rigor's sake). 

As I said, I believe the n-dimensional (at least 4-dimensional) implementation of their geomedian sits somewhere at Geoscience Australia already developed, as I can see a `geomedian_pcm` (and even a `nangeomedian_pcm`) called [here,](https://github.com/GeoscienceAustralia/burn-mapping/pull/7/files#diff-4b64c64388aca506a037e44b5433d786R17)

That's why I'm reluctant to make a PR on `hdmedians` as I'm pretty certain the work's already been done.

Paper for reference: [https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=8004469](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=8004469)

kylebarron: That seems to import `hdstats`, which isn't on PyPI... I can't find anything about that package

wouellette: I got confirmation from [Dale Roberts](https://github.com/daleroberts) that `hdstats` will be made available through PyPI in the coming months so we can stall this for now.

rio_tiler/mosaic/methods/defaults.py
wouellette: It makes a lot of sense.

The reason I didn't go for a shipped selection band as part of the asset is in the case of feeding STAC assets directly to `mosaic_reader` from an S3 bucket, this selection criterium band is not available, and that would require computing them upfront. 

I guess this could be done in the `tiler` method passed to `mosaic_reader`, as in example below (pseudo-code to illustrate the idea, I haven't tested it):

```
from satsearch import Search
from rio_tiler.mosaic import mosaic_reader
from rio_tiler.mosaic.methods import defaults
from rasterio.crs import CRS

def tiler(src_path: str, *args, **kwargs) -> Tuple[np.ndarray, np.ndarray]:
    with S2COGReader(src_path) as cog:
        cog_src =  cog.part(*args, **kwargs)
        ndvi = (cog_src[-1] - cog_src[-2]) / (cog_src[-1] - cog_src[-2])
    return np.concatenate([cog_src, ndvi], axis=0)

search = Search(bbox=bbox,
               url="https://earth-search.aws.element84.com/v0",
               datetime=f'{time_interval[0]}/{time_interval[1]}',
               query={'eo:cloud_cover': {'lt': 90}},
               collections=['sentinel-s2-l2a-cogs'])

sentinel, assets_used = mosaic_reader(assets, tiler, bbox,
                                      bounds_crs=CRS.from_epsg(bbox.crs),
                                      pixel_selection=defaults.CustomLastBandHigh(),
                                      bands=('B02', 'B03', 'B04', 'B08'))
```

The problem is that we lose the string reference to the bands in the `tiler` method, which is not really an issue as it is custom set by the user, so it can always be aligned with the bands set in `mosaic_reader`.

vincentsarago: @wouellette I think you could use `expression` for what you are trying to do.

```python
from satsearch import Search
from rio_tiler_pds.sentinel.aws import S2COGReader
from rio_tiler.mosaic import mosaic_reader
from rio_tiler.mosaic.methods import defaults
from rasterio.crs import CRS

def tiler(src_path: str, *args, **kwargs) -> Tuple[np.ndarray, np.ndarray]:
    with S2COGReader(src_path) as cog:
        return cog.part(*args, **kwargs)

search = Search(
    bbox=bbox,
    url="https://earth-search.aws.element84.com/v0",
    datetime=f'{time_interval[0]}/{time_interval[1]}',
    query={'eo:cloud_cover': {'lt': 90}},
    collections=['sentinel-s2-l2a-cogs']
)

(data, mask), assets_used = mosaic_reader(
    assets, tiler, bbox,
    bounds_crs=CRS.from_epsg(bbox.crs),
    pixel_selection=defaults.CustomLastBandHigh(),
    expression="B02,B03,B04,(B04-B08/B04+B08)"
)

# Note: data would be a (3xnxn) array of type float because of the NDVI band
```

wouellette: I completely overlooked the fact that an expression field was available...

That's sorted then! What's the plan with this? Crystallize these methods in the library? Or was your idea to keep them separate.

vincentsarago:
@wouellette I'm going to close this PR for now. Feel free to re-open when things move on hdstats

PR #271, opened 2020-10-01
remove ContextManager requirement

vincentsarago:
Started in #265, this PR removes the ContextManager requirement in the base classes. 

We also move all properties definiton from `__enter__` to `__attrs_post_init__`.

ref: https://github.com/cogeotiff/rio-tiler-pds/issues/21

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -106,7 +106,7 @@ class COGReader(BaseReader):
     _kwargs: Dict[str, Any] = attr.ib(init=False, factory=dict)
 
     def __attrs_post_init__(self):
-        """Define _kwargs."""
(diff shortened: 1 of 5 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -131,6 +129,13 @@ def __enter__(self):
...
+    def close(self):
+        """Close rasterio dataset."""
+        if self.filepath:
+            self.dataset.close()
+
+    def __enter__(self):
+        """Support using with Context Managers."""
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
```
geospatial-jeff, line 141: use `self.close()` here

geospatial-jeff (approved):

kylebarron (approved):

PR #273, opened 2020-10-02
add base classes from rio-tiler-pds

vincentsarago:
This PR moves base classes from rio-tiler-pds to rio-tiler

ref: https://github.com/cogeotiff/rio-tiler-pds/issues/24

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 import attr
 import numpy
 
-from ..errors import ExpressionMixingWarning, MissingAssets
+from ..errors import ExpressionMixingWarning, MissingAssets, MissingBands
 from ..expression import apply_expression
 from ..tasks import multi_arrays, multi_values
 
@@ -88,7 +88,6 @@ class MultiBaseReader(BaseReader, metaclass=abc.ABCMeta):
 
     reader: Type[BaseReader] = attr.ib()
     reader_options: Dict = attr.ib(factory=dict)
-    bounds: Tuple[float, float, float, float] = attr.ib(init=False)
     assets: Sequence[str] = attr.ib(init=False)
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


rio_tiler/io/base.py
```diff
@@ -21,6 +21,14 @@ class BaseReader(metaclass=abc.ABCMeta):
...
     maxzoom: int = attr.ib(init=False)
 
+    def __enter__(self):
+        """Support using with Context Managers."""
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Support using with Context Managers."""
+        pass
+
```
vincentsarago, line 31: @geospatial-jeff I know we agreed to removes those (done in https://github.com/cogeotiff/rio-tiler/pull/271/) But after some though I think it's not a great Idea. 

The main reason to remove those was to ease the use for Async but I do feel using BaseReader is not the way forward for this and we should create a `AsyncBaseReader`). 

The major problem by not having `ctx` has a requirement is that a user could create a Reader without them but then it won't be able to use it in most of our other project that use `with reader() as ...`

kylebarron (approved):

rio_tiler/io/base.py
geospatial-jeff: There is a separate `AsyncBaseReader` [here](https://github.com/cogeotiff/rio-tiler/pull/265/files#diff-2b49975a55ad278a0b6c2f8abaedc733R87).  Thinking through this again I agree that these should be here.  Its standard to have post init logic in `__init__` and keep context managers simple.  But because there is no way to execute asynchronous code within `__init__` we have to run post init logic in `__aenter__` and `__aexit__` for the async base class.

So if we are requiring the async base to be a context manager the sync base should also be to keep the interfaces consistent.  I guess there is one difference; the async base won't be usable outside the context manager, but we could provide a class method which people could call:

```python
reader = await AsyncBaseReader.open()
await reader.close()
```

vincentsarago: 🙏 thanks @geospatial-jeff

PR #279, opened 2020-10-03
Try to fix error when threads=0

kylebarron:

PR #278, opened 2020-10-03
[WIP] Mosaic raises error with TileOutsideBounds when threads <= 1

vincentsarago:
- [x] add tests
- [ ] fix 

https://github.com/cogeotiff/rio-tiler/blob/c11db1b43cbf586f4c874a4930f2d61d313ea598/rio_tiler/tasks.py#L39-L47

cc @kylebarron

PR #280, opened 2020-10-03
refactor tasks

geospatial-jeff:
Refactor `tasks.py`, mostly to differentiate between single threaded and multi threaded processing.

Started in #279

### DIFF
--- a/rio_tiler/tasks.py
+++ b/rio_tiler/tasks.py
@@ -117,7 +117,7 @@ class MultiThreadedManager(TaskManager):
                 logger.info(err)
 
 
-manager: TaskManager = TaskManager.create_from_env()
+task_manager: TaskManager = TaskManager.create_from_env()
 
 
 def multi_arrays(
@@ -128,8 +128,8 @@ def multi_arrays(
     **kwargs: Any,
 ) -> Tuple[numpy.ndarray, numpy.ndarray]:
     """Multi array."""
-    tasks = manager.create_tasks(reader, assets, *args, threads=threads, **kwargs)
-    data, masks = zip(*[r for r, _ in manager.filter_tasks(tasks)])
+    tasks = task_manager.create_tasks(reader, assets, *args, threads=threads, **kwargs)
+    data, masks = zip(*[r for r, _ in task_manager.filter_tasks(tasks)])
     data = numpy.concatenate(data)
     mask = numpy.all(masks, axis=0).astype(numpy.uint8) * 255
     return data, mask
@@ -143,5 +143,5 @@ def multi_values(
     **kwargs: Any,
 ) -> Dict:
     """Multi Values."""
-    tasks = manager.create_tasks(reader, assets, *args, threads=threads, **kwargs)
-    return {asset: val for val, asset in manager.filter_tasks(tasks)}
+    tasks = task_manager.create_tasks(reader, assets, *args, threads=threads, **kwargs)
+    return {asset: val for val, asset in task_manager.filter_tasks(tasks)}
--- a/rio_tiler/tasks.py
+++ b/rio_tiler/tasks.py
@@ -1,86 +1,138 @@
 """rio_tiler.tasks: tools for handling rio-tiler's future tasks."""
-
+import abc
 import logging
 from concurrent import futures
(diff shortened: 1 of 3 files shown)


rio_tiler/tasks.py
```diff
@@ -1,86 +1,138 @@
...
+class MultiThreadedManager(TaskManager):
+    """Multi threaded task management"""
+
+    max_threads: int
+
+    def create_tasks(
+        self, reader: Callable, assets, *args, **kwargs
+    ) -> Sequence[Tuple[futures.Future, str]]:
+        """Create Future Tasks."""
+        with futures.ThreadPoolExecutor(max_workers=self.max_threads) as executor:
```
geospatial-jeff, line 88: executor could be defined as a ClassVar if we wanted to re-use the same executor

rio_tiler/tasks.py
```diff
@@ -1,86 +1,138 @@
...
+
+        """
+        for future, asset in tasks:
+            try:
+                yield future.result(), asset
+            except allowed_exceptions as err:
+                logger.info(err)
+
+
+manager = TaskManager.create_from_env()
```
geospatial-jeff, line 119: You can monkeypatch this with your own `TaskManager`

kylebarron (approved):
Ideally would be nice to have a `threads` argument instead of only an environment variable, but otherwise looks good

vincentsarago:
@geospatial-jeff can you add tests we added in https://github.com/cogeotiff/rio-tiler/pull/278

kylebarron (approved):

vincentsarago:
FYI this is going to break https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/base.py#L13 ;-)

rio_tiler/tasks.py
```diff
@@ -47,7 +51,10 @@ def create_tasks(
         self, reader: Callable, assets, *args, **kwargs
     ) -> Generator[Tuple[Callable, str], None, None]:
         """Create Future Tasks."""
-        return ((reader(asset, *args, **kwargs), asset) for asset in assets)
+        return (
+            (reader(asset, *args, **self._filter_kwargs(kwargs)), asset)
```
kylebarron, line 55: Instead of filtering out a `threads` key from `kwargs`, would it be easier to just add `threads` as an argument to `create_tasks`?

rio_tiler/tasks.py
```diff
@@ -47,7 +51,10 @@ def create_tasks(
         self, reader: Callable, assets, *args, **kwargs
```
vincentsarago, line 51: can't we do `self, reader: Callable, assets, *args, threads=None, **kwargs` here?

rio_tiler/tasks.py
```diff
@@ -48,13 +44,10 @@ class SingleThreadedManager(TaskManager):
     max_threads: int = 1
 
     def create_tasks(
-        self, reader: Callable, assets, *args, **kwargs
+        self, reader: Callable, assets, *args, threads: int = None, **kwargs
```
kylebarron, line 47: Do we require `Optional[int]`?

rio_tiler/tasks.py
```diff
@@ -1,62 +1,123 @@
...
+@attr.s
+class TaskManager(abc.ABC):
+    """Task manager base class"""
+
+    max_threads: int = attr.ib()
+
+    @classmethod
+    def create_from_env(cls) -> "TaskManager":
+        """Create from environment."""
+        if MAX_THREADS == 1:
```
vincentsarago, line 25: this should be `MAX_THREADS =< 1:`

kylebarron:
Last thing, it might be nice to rename `manager` as `task_manager`? Then it might be a clearer import:

```py
from rio_tiler.tasks import task_manager
```
instead of
```py
from rio_tiler.tasks import manager
```

vincentsarago:
@geospatial-jeff while I like your implemtation with `TaskManager` class, this is a breaking change and maybe a bit over complicated. I pushed https://github.com/cogeotiff/rio-tiler/tree/tasksButWithWhile with a fix (because right now this version is failing), but I think we should go with a non-breaking simple solution proposed in #281

PR #281, opened 2020-10-05
wrap tasks in a while loop to avoid exceptions outside the Try/Except block

vincentsarago:


rio_tiler/io/cogeo.py
```diff
@@ -301,7 +301,7 @@ def tile(
         tile = mercantile.Tile(x=tile_x, y=tile_y, z=tile_z)
         if not tile_exists(self.bounds, tile_z, tile_x, tile_y):
             raise TileOutsideBounds(
-                "Tile {}/{}/{} is outside image bounds".format(tile_z, tile_x, tile_y)
+                f"Tile {tile_z}/{tile_x}/{tile_y} is outside {self.filepath} bounds"
```
vincentsarago, line 304: add a bit more info on this exception

rio_tiler/tasks.py
```diff
@@ -36,12 +46,15 @@ def filter_tasks(
     if not allowed_exceptions:
         allowed_exceptions = ()
 
-    for future, asset in tasks:
+    while True:
         try:
+            future, asset = next(tasks)  # type: ignore
```
vincentsarago, line 51: Not sure why mypy wasn't happy here

rio_tiler/tasks.py
```diff
@@ -36,12 +46,15 @@ def filter_tasks(
...
             if isinstance(future, futures.Future):
                 yield future.result(), asset
             else:
                 yield future, asset
+        except StopIteration:
```
vincentsarago, line 56: we put StopIteration first, so someone doesn't try to put StopIteration as `allowed_exceptions`

rio_tiler/tasks.py
```diff
@@ -51,10 +64,12 @@ def create_tasks(reader: Callable, assets, threads, *args, **kwargs) -> TaskType
     """Create Future Tasks."""
     if threads and threads > 1:
         with futures.ThreadPoolExecutor(max_workers=threads) as executor:
-            return [
-                (executor.submit(reader, asset, *args, **kwargs), asset)
-                for asset in assets
-            ]
+            return iter(
```
vincentsarago, line 67: wrapping the futures inside an iterator to be able to use `next`

rio_tiler/io/cogeo.py
kylebarron: You know I love my f-strings 😉

rio_tiler/tasks.py
kylebarron: They might wonder why their endless loop is slow...

rio_tiler/tasks.py
kylebarron: You might just be able to use a generator comprehension instead? Replace the `[ ]` with parentheses?

rio_tiler/tasks.py
vincentsarago: well if they do that (putting StopIteration in allowed_exception) the while loop while never exit ...

rio_tiler/tasks.py
vincentsarago: tried and it didn't work 🤷‍♂️

vincentsarago: (I think because the future needs to be executed with the excetutor. if you use a generator, the code will be run outide (in the filer function)

rio_tiler/tasks.py
kylebarron: haha yes, hence the endless loop will be slow 😄

rio_tiler/tasks.py
kylebarron: That makes sense 👍

PR #285, opened 2020-10-07
add support for multiple TMS via Morecantile

### LINKED ISSUE #283: Add Morecantile as dependency?
because @kylebarron loves mono-repo we could add `morecantile` as an optional dependency and archive rio-tiler-crs. 

To make sure we keep compatibily, we will need to create an Object for mercantile that replicates morecantile methods.

This shouldn't be a breaking changes.

vincentsarago:
closes #283 

This PR does:
- add a new `tms` submodule with a mock class if morecantile is not present.
- add `tms` options in Readers (**breaking change**) 
- add tile_bounds vs bounds check in `tile` methods for MultiBands and MultiBase classes
- add `tile_exists` method in BaseReader (take tms in account)
- adapt `zooms` calculation in COGReader
- deprecated `rio_tiler.reader.tile`, `rio_tiler.utils.geotiff_options`, `rio_tiler.utils.tile_exists` functions

Also does: 
- fix min zoom level calculation (as in https://github.com/cogeotiff/rio-cogeo/pull/152)


vincentsarago:
To be honest we could wait for 2.1.0 to add this ... 🤷‍♂️

rio_tiler/io/cogeo.py
```diff
@@ -5,23 +5,25 @@
...
 from .. import constants, reader
-from ..errors import ExpressionMixingWarning, TileOutsideBounds
+from ..errors import DeprecationWarning, ExpressionMixingWarning, TileOutsideBounds
 from ..expression import apply_expression, parse_expression
-from ..mercator import get_zooms
-from ..utils import has_alpha_band, has_mask_band, tile_exists
+from ..utils import has_alpha_band, has_mask_band
 from .base import BaseReader
 
+default_tms = morecantile.tms.get("WebMercatorQuad")
```
geospatial-jeff, line 25: maybe define this constant only once?  Its also in `base`

geospatial-jeff (approved):

PR #291, opened 2020-10-22
second but now tested Tasks generator with/without threads

vincentsarago:
re-do of https://github.com/cogeotiff/rio-tiler/pull/281

1. tests were poorly written
2. everything was fine when running with `1 thread` because of a `bug`
3. Generator lose its state at the first exception, meaning that we can't really go through all the stack of tasks, it's was just luck that we someone had a first COG returning (yield) before a task raise an exception.

### DIFF
--- a/rio_tiler/mosaic/reader.py
+++ b/rio_tiler/mosaic/reader.py
@@ -73,7 +73,7 @@ def mosaic_reader(
         )
 
     if not chunk_size:
-        chunk_size = threads or len(assets)
+        chunk_size = threads if threads > 1 else len(assets)
 
     assets_used: List[str] = []
 
--- a/rio_tiler/tasks.py
+++ b/rio_tiler/tasks.py
@@ -1,6 +1,7 @@
 """rio_tiler.tasks: tools for handling rio-tiler's future tasks."""
 
 from concurrent import futures
+from functools import partial
 from typing import Any, Callable, Dict, Generator, Optional, Sequence, Tuple, Union
 
 import numpy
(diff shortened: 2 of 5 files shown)


rio_tiler/mosaic/reader.py
```diff
@@ -73,7 +73,7 @@ def reader(asset: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
         )
 
     if not chunk_size:
-        chunk_size = threads or len(assets)
+        chunk_size = threads if threads > 1 else len(assets)
```
vincentsarago, line 76: this explains why it was fine with `threads=1`, because we were processing chunks of one element.

tests/test_mosaic.py
```diff
@@ -222,18 +221,64 @@ def test_mosaic_tiler_Stdev():
 
 def test_threads():
     """Test mosaic tiler."""
-    assets = [asset1, asset2, asset1, asset2, asset1, asset2]
+    assets = [asset2, asset1, asset1, asset2, asset1, asset2]
```
vincentsarago, line 224: Simple but important changes, before we were just lucky because the tiler was first yielding result for asset1. This change was made during testing to validate the new implementation

rio_tiler/tasks.py
```diff
@@ -1,26 +1,29 @@
...
 import numpy
 
 from .constants import MAX_THREADS
 from .logger import logger
 
-TaskType = Union[
-    Generator[Tuple[Callable, str], None, None], Iterator[Tuple[futures.Future, str]],
-]
+
+class LikeFuture:
```
vincentsarago, line 12: I'm open to suggestions for a better class name. 

as the name tell, this Object is like a future, meaning that it has a `.result` method which will be in a non-threading env to run the function and get the results

rio_tiler/tasks.py
```diff
@@ -44,15 +47,9 @@ def filter_tasks(
     if not allowed_exceptions:
         allowed_exceptions = ()
 
-    while True:
+    for (future, asset) in tasks:
```
vincentsarago, line 50: because we don't need generators anymore we can switch back to simple Sequence

kylebarron (approved):

PR #294, opened 2020-10-25
surface allowed_exceptions in mosaic_reader

### LINKED ISSUE #293: surface allowed_exception in mosaic_reader options
https://github.com/cogeotiff/rio-tiler/blob/62bdf186186eaee5b8294cb208d9894bd82495c9/rio_tiler/mosaic/reader.py#L83

COuld be useful if there is one file that is 
- missing 
- corrupted ...

vincentsarago:
closes #293 

FYI, we needed this because some items in the sentinel-2 stac API are not present, so we need to add `botocore.errorfactory.NoSuchKey` to the allowed exceptions.

### DIFF
--- a/rio_tiler/mosaic/reader.py
+++ b/rio_tiler/mosaic/reader.py
@@ -53,7 +53,8 @@ def mosaic_reader(
         loop. By default reads from the MAX_THREADS environment variable, and if
         not found defaults to multiprocessing.cpu_count() * 5.
     allowed_exceptions: Tuple, optional
-        List of exceptions which will be ignored.
+        List of exceptions which will be ignored. Default is set to `(TileOutsideBounds,)`.
+        Note: `TileOutsideBounds` is likely to be raised and should be included in the allowed_exceptions.
     kwargs: dict, optional
         tiler specific options.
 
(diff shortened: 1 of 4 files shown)


rio_tiler/mosaic/reader.py
```diff
@@ -80,7 +83,7 @@ def reader(asset: str, *args, **kwargs) -> Tuple[numpy.ndarray, numpy.ndarray]:
     for chunks in _chunks(assets, chunk_size):
         tasks = create_tasks(reader, chunks, threads, *args, **kwargs)
         for (t, m), asset in filter_tasks(
-            tasks, allowed_exceptions=(TileOutsideBounds,)
+            tasks, allowed_exceptions=allowed_exceptions,
```
kylebarron, line 86: A good question is whether we should always include `TileOutsideBounds`. A user who doesn't know the internals of the API might provide `allowed_exceptions=(botocore.NoSuchKey)` and then be confused why tons of new errors are coming up from `TileOutsideBounds`.

E.g. should `allowed_exceptions` be _added_ to a default of `TileOutsideBounds`?

geospatial-jeff: I think the most intuitive default behavior is to reraise any exception not in `allowed_exceptions`, having a default set of exceptions that are always makes this more confusing.

kylebarron: That's fair. Probably good to add this to the docstring that if you provide a custom list you need to include `TileOutsideBounds` explicitly if desired.

PR #295, opened 2020-10-25
add spatialinfoMixin

vincentsarago:
Following comments in #265 

this is a breaking change

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 import attr
 import numpy
 import rasterio
-from morecantile import TileMatrixSet
+from morecantile import Tile, TileMatrixSet
 from rasterio import transform
 from rasterio.crs import CRS
 from rasterio.enums import Resampling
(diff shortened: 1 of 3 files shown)


kylebarron:
I'm not against refactoring, but I think the mixin should be applied to the `BaseReader` and not the descendants of the `BaseReader`.

Because we’re not saying the fields _shouldn’t_ be there, we’re just saying the fields should be in _both_ the `BaseReader` and the `AsyncBaseReader`.

geospatial-jeff (approved):

rio_tiler/io/stac.py
```diff
@@ -76,7 +76,7 @@ def _get_assets(
 
 
 @attr.s
-class STACReader(MultiBaseReader):
+class STACReader(MultiBaseReader, SpatialMixin):
```
kylebarron, line 79: Is this necessary? The `MultiBaseReader` inherits from `BaseReader`, which already contains the `SpatialMixin`, right?

rio_tiler/io/cogeo.py
```diff
@@ -20,11 +20,11 @@
 from ..errors import DeprecationWarning, ExpressionMixingWarning, TileOutsideBounds
 from ..expression import apply_expression, parse_expression
 from ..utils import has_alpha_band, has_mask_band
-from .base import BaseReader
+from .base import BaseReader, SpatialMixin
 
 
 @attr.s
-class COGReader(BaseReader):
+class COGReader(BaseReader, SpatialMixin):
```
kylebarron, line 27: Likewise isn't this not necessary because the mixin is in the `BaseReader` already?

PR #297, opened 2020-10-30
Use pydantic Model for outputs in Readers

### LINKED ISSUE #296: add Pydantic model for metadata/info Reader output?
We now have morecantile as requirements, meaning we also have pydantic!

I think it will make sense to add output model to BaseReader.metadata or BaseReader.info. Base class don't enforce output type but I think it will still be a good thing.

https://github.com/developmentseed/titiler/blob/fe954aaf2622276b9d98f7b308a7324aabbc70be/titiler/models/dataset.py#L10-L54

vincentsarago:
closes #296 

# To Do
- [x] more changelog on deprecation/breaking changes
- [x] validate names of models
- [x] reviews

**breaking changes**
- `pc` -> `percentiles` in `stats` results 
- use `band{index}` instead of {index} in statistics Dict: `{band1: {percentiles: ....}}`: enable consistency for key type between Readers.
- band_descriptions and band_metadata are in form of `Tuple[str, Any]` (instead of Tuple[int, Any]`) with the first element being `band{index}` as in statistic dict.


rio_tiler/io/base.py
```diff
@@ -199,9 +202,9 @@ def parse_expression(self, expression: str) -> Tuple:
         _re = re.compile(assets.replace("\\\\", "\\"))
         return tuple(set(re.findall(_re, expression)))
 
-    def info(
+    def info(  # type: ignore
```
vincentsarago, line 205: we need the `type: ignore` here because the output is `Dict[str, Info]` while in the baseclass it's `Info` only

rio_tiler/io/base.py
```diff
@@ -516,7 +523,9 @@ def stats(
         def _reader(band: str, *args, **kwargs) -> Dict:
             url = self._get_band_url(band)
             with self.reader(url, tms=self.tms, **self.reader_options) as cog:  # type: ignore
-                return cog.stats(*args, **kwargs)[1]
+                return cog.stats(*args, **kwargs)[
+                    "band1"
+                ]  # We only return statistics for Band `1` of each dataset.
```
vincentsarago, line 528: put the comment before to avoid breaking line

rio_tiler/io/cogeo.py
```diff
@@ -514,126 +506,6 @@ def point(
...
-    )
-
-    def _worker(asset: str) -> Dict:
-        with COGReader(asset) as cog:
-            return cog.metadata(*args, **kwargs)
-
-    with futures.ThreadPoolExecutor(max_workers=constants.MAX_THREADS) as executor:
-        return list(executor.map(_worker, assets))
-
-
```
vincentsarago, line 636: removed deprecated functions

rio_tiler/reader.py
```diff
@@ -551,16 +553,15 @@ def metadata(
...
         for b in range(data.shape[0])
     }
 
     def _get_descr(ix):
         """Return band description."""
-        name = src_dst.descriptions[ix - 1]
-        if not name:
-            name = "band{}".format(ix)
-        return name
+        return src_dst.descriptions[ix - 1] or ""
```
vincentsarago, line 564: we do not create a band description now

rio_tiler/utils.py
```diff
@@ -82,13 +82,15 @@ def _stats(
...
     sample, edges = numpy.histogram(arr[~arr.mask], **kwargs)
-    return {
-        "pc": numpy.percentile(arr[~arr.mask], percentiles).astype(arr.dtype).tolist(),
-        "min": arr.min().item(),
-        "max": arr.max().item(),
-        "std": arr.std().item(),
-        "histogram": [sample.tolist(), edges.tolist()],
-    }
+    return dict(
+        percentiles=numpy.percentile(arr[~arr.mask], percentiles)
```
vincentsarago, line 86: Note: `pc -> percentiles`

rio_tiler/reader.py
```diff
@@ -474,7 +474,9 @@ def stats(
 
     hist_options = hist_options or {}
     return {
-        indexes[b]: raster_stats(data[b], percentiles=percentiles, **hist_options)
+        f"band{indexes[b]}": raster_stats(
```
vincentsarago, line 477: small but important breaking change! 

we now use only `str` as statistics key value, `{index} -> "band{index}"`

rio_tiler/models.py
```diff
@@ -67,3 +71,77 @@ class Metadata(Info):
...
     statistics: Dict[str, ImageStatistics]
+
+
+@attr.s
+class ImageData:
+    """Image Data class."""
+
+    data: numpy.ndarray = attr.ib()
+    mask: numpy.ndarray = attr.ib()
+    assets: Optional[List[str]] = attr.ib(default=None)
```
vincentsarago, line 82: we can add more things here!

rio_tiler/models.py
```diff
@@ -67,3 +71,77 @@ class Metadata(Info):
...
+                        self.data[bdx], in_range=rescale_arr[bdx], out_range=out_range
+                    ),
+                    0,
+                )
+            self.data = self.data.astype(numpy.uint8)
+
+        if color_formula:
+            self.data[self.data < 0] = 0
+            for ops in parse_operations(color_formula):
+                self.data = scale_dtype(ops(to_math_type(self.data)), numpy.uint8)
```
vincentsarago, line 139: This adds rio-color dependency

rio_tiler/utils.py
```diff
@@ -274,6 +274,19 @@ def linear_rescale(
...
+            f"No input range set, using image min/max values ({in_range[0], in_range[1]})",
+            UserWarning,
+        )
+
+    if out_range is None:
+        warnings.warn(
+            "No output range set, using image (0, 255)", UserWarning,
+        )
+        out_range = (0, 255)
+
```
vincentsarago, line 289: Needed update! 

- we now don't set default input range to 0,1 but to min,max of the input image! 
- The output default range is also `fixed` to be `0, 255` not `1,255`

CHANGES.md
```diff
@@ -1,4 +1,73 @@
 
+## 2.0.0rc1 (TBD)
+
+* added `ImageData` output class for all `rio_tiler.io` classed returning numpy array like types (`tile, mask = method()`)
```
kylebarron, line 4:
```suggestion
* added `ImageData` output class for all `rio_tiler.io` classes returning numpy array-like types (`tile, mask = method()`)
```

rio_tiler/constants.py
```diff
@@ -2,10 +2,15 @@
 
 import multiprocessing
 import os
+from typing import Tuple, Union
 
 import morecantile
 from rasterio.crs import CRS
 
+NumType = Union[float, int]
+BBox = Tuple[NumType, NumType, NumType, NumType]
```
kylebarron, line 11: Is it important to also let these be `int`?

rio_tiler/reader.py
kylebarron: Why is this important/necessary? Like I could see someone wanting to keep the band key as an integer so that they can then reference the original image again?

rio_tiler/utils.py
kylebarron: You love `dict()` instead of `{}` 😄

rio_tiler/utils.py
kylebarron: You could also just require the input range to be provided...? Because the local min/max isn't the same as the global min/max

rio_tiler/models.py
kylebarron: Is this the only thing we're using from `rio-color`? I'm wary of adding so many new dependencies unless really necessary. For example some people put in work to make a conda package for rio-tiler 1.x. It'll be much harder to make a Conda package for rio-tiler 2.x ...; `rio-color` isn't on Conda

kylebarron (approved):

rio_tiler/reader.py
vincentsarago: it's better (IMO) for consistency for all the readers (e.g the multiBands reader use `band` B01 as key). 

also for pydantic model to support this we need to define the key type as Union[int, str] but this is risky because a string `"1"` will be casted to int ... having just one type (str) is mush safer. 

the idea is also to have a match with the band_description, band_metadata tuples `[("band1", "{band descr}" ]`.


When building frontend using output from `metadata` or `info`, having some consistency in the key type is also better.

rio_tiler/models.py
vincentsarago: I though adding rio-tiler (color_formula) could be nice (to remove downstream code in titiler or other dynamic tiler using rio-tiler).

about the conda problem ... well 🤷‍♂️

rio_tiler/utils.py
vincentsarago: yes, I was worry about breaking more this function because it always had defaults. 

about min/max, the linear rescale could be use on `preview` which would be a valid case where to use min/max of the array.

rio_tiler/constants.py
vincentsarago: I don't know 🤷‍♂️

rio_tiler/models.py
```diff
@@ -0,0 +1,69 @@
...
+    Empty = "None"
+
+
+class Bounds(BaseModel):
+    """Dataset Bounding box"""
+
+    bounds: BBox
+
+
+class SpatialInfo(Bounds):
```
geospatial-jeff, line 29: Do we want to make these dict-like to be backwards compatible?  Similar to the tile responses.

vincentsarago: yes it will be better... I assume it wasn't possible but didn't check :D

vincentsarago: @geospatial-jeff can you check at https://github.com/cogeotiff/rio-tiler/pull/297/commits/dda29be27d249d0e644cc23125280e34facebe94 to see if I'm not doing something stupid? 😄

rio_tiler/reader.py
vincentsarago: well In fact I'm going to re-change this AGAIN 🤦.
Setting for `band{}` each info (stats, band_description, band_metadata) doesn't help when you are working with simple COG (via COGReader) or Multiple Bands (via MultiBandsReader). 

For COG Reader we need to be able to set an `indexes=1,2,3,...` options while for MultiBands we need `bands=a,b,c`. The `key` should help to construct those.

Still we will enforce the key to be a string instead of an integer.

rio_tiler/models.py
```diff
@@ -107,10 +107,10 @@ def create_from_list(cls, data: Sequence["ImageData"]):
         assets = [img.assets[0] for img in data if img.assets]
 
         bounds_values = [img.bounds for img in data if img.bounds]
-        bounds = list(set(bounds_values))[0] if bounds_values else None
+        bounds = bounds_values[0] if bounds_values else None
```
vincentsarago, line 110: we don't really need Set here. Also in some `List` is not hashable.

rio_tiler/models.py
```diff
@@ -155,10 +155,27 @@ def transform(self):
     def post_process(
         self,
         in_range: Optional[Tuple[NumType, NumType]] = None,
-        out_range: Optional[Tuple[NumType, NumType]] = None,
+        out_dtype: str = "uint8",
```
kylebarron, line 158: For typing I think you should allow more than just `str`. I.e. I prefer to use `np.uint8` for typing, and that would fail here. So I'd suggest

```suggestion
        out_dtype: Union[str, numpy.number] = "uint8",
```

`np.number` should cover all float and integer types

rio_tiler/models.py
```diff
@@ -7,13 +7,14 @@
 import numpy
 from affine import Affine
 from pydantic import BaseModel
+from rasterio.coords import BoundingBox
```
kylebarron, line 10: ❤️

PR #307, opened 2020-11-24
add feature method

### LINKED ISSUE #306: Add method to replicate https://cogeotiff.github.io/rio-tiler/advanced/feature/
Reading a file for a specific feature seems a bit more natural than using a bbox for some users

We have a way of doing this using `cutline`

```python
from rio_tiler.io import COGReader
from rio_tiler.utils import create_cutline
from rasterio.features import bounds as featureBounds

feat =     {
    "type": "Feature",
    "properties": {},
    "geometry": {
    "type": "Polygon",
    "coordinates": [
        [
        [-52.6025390625, 73.86761239709705],
        [-52.6025390625, 73.59679245247814],
        [-51.591796875, 73.60299628304274],
        [-51.591796875, 73.90420357134279],
        [-52.4267578125, 74.0437225981325],
        [-52.6025390625, 73.86761239709705]
        ]
    ]
    }
}

# Get BBOX of the polygon
bbox = featureBounds(feat)

# Use COGReader to open and read the dataset
with COGReader("my_tif.tif") as cog:
    # Create WTT Cutline
    cutline = create_cutline(cog.dataset, feat, geometry_crs="epsg:4326")

    # Read part of the data (bbox) and use the cutline to mask the data
    data, mask = cog.part(bbox, vrt_options={'cutline': cutline})
```

but it will be nicer if we could directly pass the feature to a method (or to the `part()` method)

```python
from rio_tiler.io import COGReader

feat =     {
    "type": "Feature",
    "properties": {},
    "geometry": {
    "type": "Polygon",
    "coordinates": [
        [
        [-52.6025390625, 73.86761239709705],
        [-52.6025390625, 73.59679245247814],
        [-51.591796875, 73.60299628304274],
        [-51.591796875, 73.90420357134279],
        [-52.4267578125, 74.0437225981325],
        [-52.6025390625, 73.86761239709705]
        ]
    ]
    }
}

with COGReader("my_tif.tif") as cog:
    data, mask = cog.part(None, feature=feat)
```

vincentsarago:
closes #306 

Question:
- are we ok supporting only `Polygon` feature (we could in theory support MultiPolygon) ?
- are we ok adding a new required method `feature` to the base class (breaking change) ?

## To Do
- add tests for Multi* and async classes
- update docs

### DIFF
--- a/docs/readers.md
+++ b/docs/readers.md
@@ -37,7 +37,7 @@ with COGReader("myfile.tif"s) as cog:
     assert img.data.count == 1
 ```
 
-- **part()**: Read part of a raster
+- **part()**: Read a raster for a given bounding box (`bbox`)
 
 ```python
 with COGReader("myfile.tif") as cog:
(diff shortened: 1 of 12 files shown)


docs/readers.md
```diff
@@ -37,7 +37,7 @@ with COGReader("myfile.tif"s) as cog:
     assert img.data.count == 1
 ```
 
-- **part()**: Read part of a raster
+- **part()**: Read a raster for a given bounding box (`bbox`)
```
kylebarron, line 40: Might be good to note that this is in image coordinates, not geospatial coordinates.

vincentsarago: > Might be good to note that this is in image coordinates, not geospatial coordinates.

well that's the opposite, `part()` takes a WGS84 bounding box, but yes 👍

rio_tiler/io/cogeo.py
```diff
@@ -520,6 +521,91 @@ def point(
...
+            )
+
+        if expression:
+            indexes = parse_expression(expression)
+
+        if not dst_crs:
+            dst_crs = shape_crs
+
+        # Get BBOX of the polygon
+        bbox = featureBounds(shape)
```
kylebarron, line 578: Do you want to raise something like a `FeatureOutsideBoundsError` when the bbox doesn't intersect image coordinates?

kylebarron (approved):

rio_tiler/io/cogeo.py
vincentsarago: I don't think we raise error like that in `.part()` method so I feel it's ok to not raise anything 🤷‍♂️ . The fact we raise an error `TileOutsideBounds` within the `.tile()` method is first historical but also because it requires more knowledge to understand if a tile is outside or outside a dataset bounds. 

Again not 💯 sold on `not throwing error`, so if you have a good argument I'm listening 😄 !

PR #309, opened 2020-12-03
add NPZ output format

### LINKED ISSUE #308: Add NPZ
https://github.com/cogeotiff/rio-tiler/blob/99d0a1616467bb9893acee54e8a96117cf097113/rio_tiler/utils.py#L438-L446

ref: https://numpy.org/doc/stable/reference/generated/numpy.savez.html#numpy.savez

vincentsarago:
closes #308


rio_tiler/utils.py
```diff
@@ -432,16 +432,28 @@ def render(
...
 
-    if img_format == "NPY":
+    elif img_format == "NPY":
         # If mask is not None we add it as the last band
         if mask is not None:
             mask = numpy.expand_dims(mask, axis=0)
             tile = numpy.concatenate((tile, mask))
         bio = BytesIO()
-        numpy.save(bio, tile)
+        numpy.save(bio, tile) if img_format == "NPY" else numpy.savez_compressed(
```
vincentsarago, line 445: 🤦

kylebarron (approved):
I think this is ok, but it might not be necessary for `rio-tiler` to support every image format natively

vincentsarago:
@kylebarron rio-tiler supports all GDAL supported image formats by default. I don't plan to add more `hacks` like NPY or NPZ, they should remain exceptions.

PR #310, opened 2020-12-05
Use generator for single-threaded task generation

kylebarron:
Edit: This PR should be a performance improvement for the `tasks` generator when using a single thread and for situations like a mosaic using `first` pixel selection, where you don't necessarily need to wait for _all_ the data returned from the generator.

-----

Original argument, since disproven with multithreaded code:

The issue is that here
https://github.com/cogeotiff/rio-tiler/blob/ff068185b269b4e556b4f922f0db6bca439763b9/rio_tiler/tasks.py#L50-L54

Since we use a list comprehension, the entire iterable will be "spent" and turned into a list before the first item will be yielded. Here's a contrived example with `sleep` to show my point:

```py
from concurrent import futures
from time import sleep

def create_tasks_map(threads=1):
    with futures.ThreadPoolExecutor(max_workers=threads) as executor:
        yield from executor.map(sleep, range(1, 4))

def create_tasks_gen(threads=1):
    with futures.ThreadPoolExecutor(max_workers=threads) as executor:
        yield from (
            executor.submit(sleep, i)
            for i in range(1, 4)
        )

def create_tasks_list(threads=1):
    with futures.ThreadPoolExecutor(max_workers=threads) as executor:
        return iter([
            executor.submit(sleep, i)
            for i in range(1, 4)
        ])

%time next(create_tasks_map())
# CPU times: user 851 µs, sys: 1.04 ms, total: 1.89 ms
# Wall time: 3 s

%time next(create_tasks_gen())
# CPU times: user 661 µs, sys: 927 µs, total: 1.59 ms
# Wall time: 1 s

%time next(create_tasks_list())
# CPU times: user 1.18 ms, sys: 1.11 ms, total: 2.29 ms
# Wall time: 6.01 s
```

Here you can see that when only loading the first item, the `yield from` is much faster because the list comprehension runs all the `sleep` calls before returning anything.

Side note, [it looks like under the hood `yield from` is pretty cool...](https://stackoverflow.com/a/26109157) 

> `yield from` ... establishes a transparent bidirectional connection between the caller and the sub-generator

### DIFF
--- a/rio_tiler/tasks.py
+++ b/rio_tiler/tasks.py
@@ -8,7 +8,7 @@ from .constants import MAX_THREADS
 from .logger import logger
 from .models import ImageData
 
-TaskType = Sequence[Tuple[Union[futures.Future, Callable], str]]
+TaskType = Generator[Tuple[Union[futures.Future, Callable], str], None, None]
 
 
 def filter_tasks(
@@ -48,13 +48,15 @@ def create_tasks(reader: Callable, assets, threads, *args, **kwargs) -> TaskType
     if threads and threads > 1:
         logger.debug(f"Running tasks in ThreadPool with max_workers={threads}")
         with futures.ThreadPoolExecutor(max_workers=threads) as executor:
-            return [
+            yield from (
                 (executor.submit(reader, asset, *args, **kwargs), asset)
                 for asset in assets
-            ]
+            )
     else:
         logger.debug(f"Running tasks outside ThreadsPool (max_workers={threads})")
-        return [(partial(reader, asset, *args, **kwargs), asset) for asset in assets]
+        yield from (
+            (partial(reader, asset, *args, **kwargs), asset) for asset in assets
+        )
 
 
 def multi_arrays(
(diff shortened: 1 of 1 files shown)


vincentsarago:
@kylebarron any update from the discussion/test over the week end ?

vincentsarago (approved):

kylebarron:
I think the conclusion from talking with @geospatial-jeff is that this is bad practice for the multithreaded case because it doesn't _send_ each item to the threadpool until the next item from the generator is called. So it's implicitly single-threaded.

But it would still be good to change the non-threaded code from a list to a generator.

vincentsarago:
@kylebarron I'm still not sure on this one. Because we use partial to trigger the call to the `reader`, I don't see how using a generator will make the code run differently 

https://github.com/cogeotiff/rio-tiler/blob/5d8dc218e4a9dc2dbbeba514f234a304fd43eacc/rio_tiler/tasks.py#L40

kylebarron:
😭 I really thought I was finding an improvement!

I didn't realize that `partial` doesn't _execute_ the function, it just wraps it, so the function isn't actually called until that `yield future(), asset` line

Issue #212, opened 2020-07-31
[STAC] resolves relative asset path


vincentsarago:
ref: https://github.com/developmentseed/titiler/issues/36#issue-639793489

from @fabricebrito 
> This feature would ease our setup as the STAC catalogues we use move from a location to another and this is why we use relative hrefs also for the asset paths. For instance pystac resolves the the full href with e.g.
```python
item.assets['nbr'].get_absolute_href()
```

vincentsarago:
it should append here https://github.com/cogeotiff/rio-tiler/blob/ce156bc5aaa52c579ee0891998eef7a97e44469c/rio_tiler/io/stac.py#L184-L190

emmanuelmathot:
Is there progress on this issue?

vincentsarago:
@emmanuelmathot no, mostly because I have no idea how to do it 🤷‍♂️, I've looked at pystac but got lost!

emmanuelmathot:
No problem, I modified the STACReader to use pystac. PR#313

PR #313, opened 2020-12-11
Using pystac for reading STAC items

emmanuelmathot:
**Proposed Changes:**

***Using pystac***
using [pystac](https://github.com/stac-utils/pystac) to manipulate stac items.
Among other features, it allows to resolve absolute url from relative links in assets. It should fix https://github.com/cogeotiff/rio-tiler/issues/212

**PR Checklist:**

### DIFF
--- a/rio_tiler/io/stac.py
+++ b/rio_tiler/io/stac.py
@@ -2,6 +2,7 @@
 
 import functools
 import json
+import pystac
 from typing import Dict, Iterator, Optional, Set, Type
 from urllib.parse import urlparse
 
@@ -45,15 +46,15 @@ def fetch(filepath: str) -> Dict:
 
 
 def _get_assets(
-    item: Dict,
+    item: pystac.Item,
     include: Optional[Set[str]] = None,
     exclude: Optional[Set[str]] = None,
     include_asset_types: Optional[Set[str]] = None,
     exclude_asset_types: Optional[Set[str]] = None,
 ) -> Iterator:
     """Get Asset list."""
-    for asset, asset_info in item["assets"].items():
-        _type = asset_info.get("type")
+    for asset, asset_info in item.get_assets().items():
+        _type = asset_info.media_type
 
         if exclude and asset in exclude:
             continue
@@ -100,8 +101,8 @@ class STACReader(MultiBaseReader):
     ----------
     filepath: str
         STAC Item path, URL or S3 URL.
-    item: Dict, optional
-        STAC Item dict.
+    item: pystac.Item, optional
+        STAC Item.
     minzoom: int, optional
         Set minzoom for the tiles.
     minzoom: int, optional
@@ -148,7 +149,7 @@ class STACReader(MultiBaseReader):
     """
 
     filepath: str = attr.ib()
-    item: Dict = attr.ib(default=None)
+    item: pystac.Item = attr.ib(default=None)
     tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
     minzoom: int = attr.ib(default=0)
     maxzoom: int = attr.ib(default=30)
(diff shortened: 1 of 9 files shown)


rio_tiler/io/stac.py
```diff
@@ -148,7 +149,7 @@ class STACReader(MultiBaseReader):
     """
 
     filepath: str = attr.ib()
-    item: Dict = attr.ib(default=None)
+    item: pystac.Item = attr.ib(default=None)
```
vincentsarago, line 152: Could we maybe use attr.converter (https://www.attrs.org/en/stable/init.html#converters) to still allow user to pass dict?

emmanuelmathot: I could but then we cannot ensure to have the self href of the item if it is not specified as a proper "self" link in the item. this self href allows to resolve relative asset uri.

emmanuelmathot: I replaced the constructure using dict. The tiny drawback is that a dictionary without proper self is unlocatable and thus relative href shall fail.

vincentsarago: I think you could do 

```python
def to_stac_item(item: Union[Dict, pystac.Item]) -> pystac.Item:
    if isinstance(item, Dict):
        return pystac.Item.from_dict(item)
    return item

item: pystac.Item = attr.ib(default=None, converter=to_stac_item)
```

emmanuelmathot: ok thank you I will do that. I am not very fond of python...

rio_tiler/io/stac.py
```diff
@@ -101,7 +109,9 @@ class STACReader(MultiBaseReader):
     filepath: str
         STAC Item path, URL or S3 URL.
     item: Dict, optional
-        STAC Item dict.
+        STAC Item as dictionary.
+    item: pystac.Item, optional
+        STAC Item.
```
vincentsarago, line 114: Duplicated info!

vincentsarago: I Think it's ok to have only
```
item: pystac.Item, optional
         STAC Item.
```

vincentsarago:
@emmanuelmathot because the goal of the PR is to fix #212 could we add a tests case for it (read a stac item with relative path and just make sure that `_get_asset_url` return the expected url.

geospatial-jeff (approved):
lgtm!

rio_tiler/io/stac.py
```diff
@@ -100,8 +108,8 @@ class STACReader(MultiBaseReader):
     ----------
     filepath: str
         STAC Item path, URL or S3 URL.
-    item: Dict, optional
-        STAC Item dict.
+    item: pystac.Item, optional
```
kylebarron, line 111: This can be `dict` or `pystac.Item`, right? Because of the `converter=to_stac_item`?

kylebarron (approved):
I think it would be good to allow both
```py
STACReader('url/to/filepath')
```
and 
```py
stac_item = {...}
STACReader(item=stac_item)
```
Though I'd guess that right now you'd have to use `STACReader(None, item=stac_item)`, because `filepath` doesn't have a default. I'd ideally change `filepath` to default to `None` and then just check in `__attrs_post_init__` that one of `filepath` and `item` exist.

rio_tiler/io/stac.py
```diff
@@ -148,7 +156,7 @@ class STACReader(MultiBaseReader):
     """
 
     filepath: str = attr.ib()
-    item: Dict = attr.ib(default=None)
+    item: pystac.Item = attr.ib(default=None, converter=to_stac_item)
```
vincentsarago, line 159: 🤔 I think `converter` will always be run. we got lucky because `None` will be returned directly (not caught by `if isinstance...`) I'm fine with this because honestly I have no better way to do it 

Example:
```python
@attr.s
class a():
    b : int = attr.ib(default=None, converter=int)

a()
> TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

vincentsarago:
thanks @emmanuelmathot 🙏 

I've made some small changes from this PR and merged it to master via another PR.

PR #316, opened 2020-12-14
delete deprecated function and add deprecation warning for rio_tiler.mercator

### LINKED ISSUE #315: deprecate rio_tiler.mercator submodule
https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/mercator.py

with the switch to morecantile, we don't need this anymore.

vincentsarago:
closes #315


kylebarron:
FYI you should use the `DeprecationWarning` from the standard library instead of making your own: https://docs.python.org/3/library/exceptions.html#DeprecationWarning

PR #324, opened 2020-12-21
add multipolygon support in cutline

### LINKED ISSUE #323: support MultiPolygon cutline
Unsure if this is supported by the underlying libs but if possible, would be good for this to accept `MultiPolygon` as well: https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L476

vincentsarago:
closes #323 

cc @drewbo

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -10,6 +10,7 @@ from boto3.session import Session as boto3_session
 from rasterio import windows
 from rasterio.crs import CRS
 from rasterio.enums import ColorInterp, MaskFlags
+from rasterio.features import is_valid_geom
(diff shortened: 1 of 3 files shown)


rio_tiler/utils.py
```diff
@@ -471,18 +472,34 @@ def create_cutline(src_dst: DataSet, geometry: Dict, geometry_crs: CRS = None) -
     if "geometry" in geometry:
         geometry = geometry["geometry"]
 
+    if not is_valid_geom(geometry):
+        raise RioTilerError("Invalid geometry")
```
vincentsarago, line 476: This PR also add geometry validation!

PR #325, opened 2020-12-22
apply discrete cmap when needed

### LINKED ISSUE #321: automatically apply `discrete` colormap when cmap doesn't have 256 values
When a user has a custom colormap which is not a 256 dict (for 0 to 256) we should automatically switch to `apply_discrete_cmap` because 
- `make_lut` won't work with dict which doesn't have 256 values
- we don't want to make a LUT with more than 256 value (e.g if a cmap is like `{0: [...], 1000:[...]}` a lut would be really large, while we could use `apply_discrete_cmap` directly.
 
https://github.com/cogeotiff/rio-tiler/blob/79da537117c9c8d70200b6b244887f530de42f20/rio_tiler/colormap.py#L97

This is to support https://github.com/developmentseed/titiler/issues/153

vincentsarago:
closes #321

### DIFF
--- a/rio_tiler/colormap.py
+++ b/rio_tiler/colormap.py
@@ -94,6 +94,9 @@ def apply_cmap(
     if data.shape[0] > 1:
         raise InvalidFormat("Source data must be 1 band")
 
+    if len(colormap) > 256 or max(colormap) > 256:
+        return apply_discrete_cmap(data, colormap)
+
     lookup_table = make_lut(colormap)
     data = lookup_table[data[0], :]
 
--- a/tests/test_cmap.py
+++ b/tests/test_cmap.py
@@ -1,5 +1,7 @@
 """tests rio_tiler colormaps"""
 
+from copy import deepcopy
+
 import numpy
 import pytest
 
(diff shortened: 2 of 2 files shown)


geospatial-jeff (approved):

rio_tiler/colormap.py
```diff
@@ -94,6 +94,9 @@ def apply_cmap(
     if data.shape[0] > 1:
         raise InvalidFormat("Source data must be 1 band")
 
+    if len(colormap) > 256 or max(colormap) > 256:
```
kylebarron, line 97: Why do you only check for `> 256` and not `!= 256`?

vincentsarago: @kylebarron because `make_lut` should work for cmap that have less than 256 value or a max smaller than 256 
https://github.com/cogeotiff/rio-tiler/pull/325/files#diff-1c0d3ee98487c3ad9e96a708cf2b17d8dba80980c2b0bd6faaca3688543bfd41R66-R68

just realizing that it should be `max(colormap) >= 256` 🤦

vincentsarago: updated the code with comments: https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/colormap.py#L97-L101

PR #333, opened 2021-01-23
set expression to be None not `""`

vincentsarago:
Setting this to None is better (if think I originally set those to `""` so solve some mypy warnings)


rio_tiler/io/base.py
```diff
@@ -485,13 +485,13 @@ def tile(
         if isinstance(assets, str):
             assets = (assets,)
 
-        if assets and expression:
+        if assets and expression is not None:
```
kylebarron, line 488: Why do you need `is not None`? Do you want `''` to be consider truthy?

kylebarron (approved):

PR #334, opened 2021-01-23
make reader.read public

### LINKED ISSUE #332: make `rio_tiler.reader.read` public
https://github.com/cogeotiff/rio-tiler/blob/962476f446efdceb94e56c7817a5633c975421e6/rio_tiler/reader.py#L24-L38

Because the function is private it is not rendered in the API docs. I think making this function public will be 👌

vincentsarago:
closes #332

### DIFF
--- a/docs/v2_migration.md
+++ b/docs/v2_migration.md
@@ -98,7 +98,7 @@ with rasterio.open("my_tif.tif") as src_dst:
     mercator_tile = mercantile.Tile(x=tile_x, y=tile_y, z=tile_z)
     tile_bounds = mercantile.xy_bounds(mercator_tile)
 
-    t, m = rio_tiler.utils._tilesrc, tile_bounds, 256)
+    t, m = rio_tiler.utils._tile_read(src, tile_bounds, 256)
 
 # v2
 with rasterio.open("my_tif.tif") as src_dst:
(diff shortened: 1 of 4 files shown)


docs/v2_migration.md
```diff
@@ -124,7 +124,7 @@ with rasterio.open("my_tif.tif") as src_dst:
 with rasterio.open("my_tif.tif") as src_dst:
     mercator_tile = mercantile.Tile(x=tile_x, y=tile_y, z=tile_z)
     tile_bounds = mercantile.xy_bounds(mercator_tile)
-    t, m = rio_tiler.utils._tile_read(src, tile_bounds, 256, tile_edge_padding=4, minimum_tile_cover=0.3)
+    t, m = rio_tiler.utils._tile_read, tile_bounds, 256, tile_edge_padding=4, minimum_tile_cover=0.3)
```
kylebarron, line 127: 1. This is definitely a typo because now it's invalid syntax
2. Both here and below you're not using `src_dst` at all, so it doesn't make sense
3. I guess this is for v1, but I'd recommend not advertising private functions.

kylebarron (approved):

docs/v2_migration.md
vincentsarago: 🤦

PR #341, opened 2021-02-04
fix docs links

chrieke:
fix 2 docs links

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,11 +117,9 @@ Create Mapbox Vector Tiles from raster sources
 
 ## Implementations
 
-#### [**rio-viz**][rio-viz]
+#### [**rio-viz**](https://github.com/developmentseed/rio-viz)
 
 ![](https://user-images.githubusercontent.com/10407788/105772356-0ca2d900-5f30-11eb-85b9-c3da9e12b663.jpg)
-
-[rio-viz]: (https://github.com/developmentseed/rio-viz)
 
 Visualize Cloud Optimized GeoTIFFs locally in the browser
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,12 +117,10 @@ Create Mapbox Vector Tiles from raster sources
 
 ## Implementations
 
-#### [**rio-viz**][rio-viz]
+#### [**rio-viz**](https://github.com/developmentseed/rio-viz)
 
 ![](https://user-images.githubusercontent.com/10407788/105772356-0ca2d900-5f30-11eb-85b9-c3da9e12b663.jpg)
 
-[rio-viz]: (https://github.com/developmentseed/rio-viz)
```
kylebarron, line 124: Oh this isn't supposed to be in parentheses. I'd suggest just removing the parentheses here instead of making an inline link.

docs/intro.md
```diff
@@ -127,6 +127,4 @@ The [Cloud-Optimized GeoTIFF (COG)](https://www.cogeo.org/) format is the **reco
 - has a header with a `map` of all the tiles
 - can have internal overviews
 
-To learn more about efficiency of COG vs other file formats, check out [this blog post](vincent_s2_jp2_cost).
```
kylebarron, line 130: Likewise, this should be

```
[this blog post][vincent_s2_jp2_cost].
```
with square quotes.

kylebarron:
I'm a fan of referenced links, since they're more readable in Markdown, but I think I suggested Vincent use them and didn't read the PR closely enough 😄

README.md
chrieke: That also works :)

kylebarron (approved):
Thanks!

vincentsarago:
🙏 thanks @chrieke and @kylebarron

PR #343, opened 2021-02-04
fix issue 342

### LINKED ISSUE #339: Update PyPi description
https://github.com/cogeotiff/rio-tiler/blob/4bd9be5db98827e40d4222627c3af59e9063bd95/setup.py#L39

Not true anymore!

### LINKED ISSUE #342: Conflicting arguments/options in mosaic_reader
https://github.com/cogeotiff/rio-tiler/blob/dd25eb8e048a2e4ff3a7a4451ab81b11332b51c7/rio_tiler/mosaic/reader.py#L18

When using the STACReader, we sometimes need to pass `assets` (or expression) to the tile/point... methods. sadly the `assets` options will be in conflict with the mosaic_reader `assets`.

let's say we have a couple of STAC items, for each one we want to read the `asset1`, and use mosaic_reader to create a tile.  we will need to use a function like:

```python
# https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/base.py#L144-L159
def tile(items, x, y, z, **kwargs) -> ImageData:

    def _reader(item: str, x: int, y: int, z: int, **kwargs: Any) -> ImageData:
        with STACReader(item) as src_dst:
            return src_dst.tile(x, y, z, **kwargs)


    return mosaic_reader(items, _reader, x, y, z, **kwargs)
```

```python
items = ["stac2.json", "stac2.json"]

img = tile(items, 1, 1, 1, assets="asset1") 
```

This will not work because mosaic_reader will have twice `assets` defined and will raise:
```
TypeError: mosaic_reader() got multiple values for argument 'assets'
```

cc @kylebarron

vincentsarago:
closes #342 , closes #339

## To Do 
- [x] add tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 
-## 2.0.0.post1 (2021-02-04)
+## 2.0.1 (2021-02-04)
 
 * fix arguments names conflicts between mosaic_reader/tasks and STACReader options (https://github.com/cogeotiff/rio-tiler/pull/343)
 * update rio-tiler pypi description.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,9 @@
 
+## 2.0.0.post1 (2021-02-04)
+
+* fix arguments names conflicts between mosaic_reader/tasks and STACReader options (https://github.com/cogeotiff/rio-tiler/pull/343)
+* update rio-tiler pypi description.
+
 ## 2.0.0 (2021-01-27)
 
 * add MultiPolygon support in `rio_tiler.utils.create_cutline` (https://github.com/cogeotiff/rio-tiler/issues/323)
--- a/tests/test_io_stac.py
+++ b/tests/test_io_stac.py
@@ -374,15 +374,12 @@ def test_feature_valid(rio):
             assert data.shape == (1, 119, 97)
 
 
-@patch("rio_tiler.io.cogeo.rasterio")
-def test_relative_assets(rio):
+def test_relative_assets():
     """Should return absolute href for assets"""
-    rio.open = mock_rasterio_open
-
     with STACReader(STAC_REL_PATH) as stac:
-
         for (key, asset) in stac.item.assets.items():
             assert asset.get_absolute_href().startswith(PREFIX)
         assert len(stac.assets) == 5
+
         for asset in stac.assets:
             assert stac._get_asset_url(asset).startswith(PREFIX)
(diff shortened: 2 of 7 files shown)


geospatial-jeff (approved):

CHANGES.md
```diff
@@ -1,4 +1,9 @@
 
+## 2.0.0.post1 (2021-02-04)
```
kylebarron, line 2: Why not just `2.0.1`? I think this counts as a bug fix...

CHANGES.md
```diff
@@ -1,4 +1,9 @@
...
+
+* fix arguments names conflicts between mosaic_reader/tasks and STACReader options (https://github.com/cogeotiff/rio-tiler/pull/343)
```
kylebarron, line 4: Tiny nit: I think putting `#343` is enough instead of the full URL

tests/test_io_stac.py
```diff
@@ -374,15 +374,12 @@ def test_feature_valid(rio):
             assert data.shape == (1, 119, 97)
 
 
-@patch("rio_tiler.io.cogeo.rasterio")
-def test_relative_assets(rio):
+def test_relative_assets():
```
kylebarron, line 377: You don't need this `@patch` anymore?

kylebarron: Ah I see it moved to `test_mosaic.py`

kylebarron (approved):

tests/test_io_stac.py
vincentsarago: well this is unrelated to the PR, I removed it because it wasn't needed, this test doesn't touch rasterio at all, just check that the links are 👌

CHANGES.md
vincentsarago: last time I tried, it didn't work, maybe it only work when deployed with Mkdocs but I think adding full link has better overall support 🤷‍♂️

CHANGES.md
vincentsarago: yeah I was a bit hesitant to make a `proper` release for just variable renames 🤷‍♂️

kylebarron: I think sticking to just `major.minor.bugfix` is simplest. Bug fix releases are "cheap" to make in my opinion

CHANGES.md
kylebarron: It should display correctly in Mkdocs: https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/#shorthand-links.

But in the first comment I wasn't even intending that. I was just saying that not-a-link is ok too, but it's really personal preference.

Issue #164, opened 2020-03-20
Bug in new `calculate_default_transform` from terracotta


vincentsarago:
While using the new `calculate_default_transform` from Terracotta seemed to help with the performance, I encountered a but when dealing with scene that crosses the dateline separation

![](https://user-images.githubusercontent.com/10407788/77132485-07cee800-6a36-11ea-8edb-a196decf25a3.png)

The problem happens here: https://github.com/cogeotiff/rio-tiler/blob/06504ec8321d185602a5a91bc6b5415bdb94915c/rio_tiler/utils.py#L232-L241

which will return `negative` width/height 

```
Tile
x=17
y=21
z=6

vrt_transfrom
|-386.76, 0.00,-9392582.04|
| 0.00, 6.31, 6887893.49|
| 0.00, 0.00, 1.00|

vrt_width=-1619
vrt_height=-99186
```

dataset link: https://www.dropbox.com/s/jyskcxqnbq3240c/ask_cog.tif?dl=0

Not sure if it means something but here is the result from both rasterio and terracotta's calculate_default_transform for a random tile.

```
# rasterio
| 238.34, 0.00,-20037508.08|
| 0.00,-238.34, 13596067.67|
| 0.00, 0.00, 1.00|

# terracotta
|-386.54, 0.00,-20028306.76|
| 0.00, 6.31, 13596067.67|
| 0.00, 0.00, 1.00|
```

@dionhaefner, could you test terracotta with the dataset I shared, so I can confirm this is only happening in rio-tiler ?

dionhaefner:
Oops, sorry! We actually have an open issue for datasets spanning the date line, since we never tested it (and I suspected it wouldn't work):

https://github.com/DHI-GRAS/terracotta/issues/53

I'll see if I can fix it, thanks for the test data.

dionhaefner:
Yikes, this image is pretty much the perfect storm for my method

<img width="501" alt="Screen Shot 2020-03-20 at 14 50 08" src="https://user-images.githubusercontent.com/11994217/77169625-19f26a00-6aba-11ea-9d48-a8001b766497.png">

- It is rotated so much that the top right corner is *further west* than the bottom left corner
- Only one corner of the raster crosses the date line
- Top left and bottom right corner of the image lie almost on the same latitude
- Bottom left and top right corner lie almost on the same longitude

[...]

PR #345, opened 2021-02-08
Correct bounds to be aligned with internal block

### LINKED ISSUE #338: Check if transform is ok
first mentioned in https://github.com/cogeotiff/rio-cogeo/issues/164

I'm seeing some weird Internal Tile fetching behavior. When working with a Web Mercator aligned file I would expect GDAL to only do 2 GET requests (header + tile) but It seems that we are fetching more than 1 tile.

```
$ tilebench profile https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256pxWEBP.tif --tile 5-10-9 --add-kernels | jq
{
  "LIST": {
    "count": 0
  },
  "HEAD": {
    "count": 1
  },
  "GET": {
    "count": 3,
    "bytes": 17144,
    "ranges": [
      "0-16383",
      "205900-206279",
      "242386-242765"
    ]
  },
  "WarpKernels": [
    [
      "Src=512,2303,257x129",
      "Dst=0,0,256x128"
    ],
    [
      "Src=512,2431,257x129",
      "Dst=0,128,256x128"
    ]
  ],
  "Timing": 0.4853549003601074
}
```

☝️ we do 3 GET, 
```
[
      "0-16383",  <-- GeoTIFF Header
      "205900-206279",  <--- Tile Data
      "242386-242765".  <--- Tile Data
]
```

The weird part is when we look at the WarpKernels:

```
[
    [
      "Src=512,2303,257x129",
      "Dst=0,0,256x128"
    ],
    [
      "Src=512,2431,257x129",
      "Dst=0,128,256x128"
    ]
]
```

I'm not sure how to read the Kernels but the `257x129` seems weird.

For the particular tile 5-10-9, here are the WarpedVRT info

```
{
    'add_alpha': True, 
    'resampling': <Resampling.nearest: 0>, 
    'crs': CRS.from_epsg(3857), 
    'transform': Affine(2445.984905125635, 0.0, -7514065.6285459455, 0.0, -2445.984905125635, 8766409.89997027), 
    'width': 512, 
    'height': 512
}
Window(col_off=0, row_off=0, width=512, height=512)
(3, 256, 256)
```

The VRT is 512x512 for a 256x256 tile, which is what we expect because the zoom 5 correspond to the first overview.

The input and output size are ok, so maybe we need to check if the `transform` is aligned with the internal tiles. 
https://github.com/cogeotiff/rio-tiler/blob/4bd9be5db98827e40d4222627c3af59e9063bd95/rio_tiler/utils.py#L166-L182

vincentsarago:
closes #338 

To be discussed

This PR adds a check to see if the requested bounds are almost aligned with the internal block. If yes then we use the bounds the block. 

```python
import rasterio
from rasterio import windows
import morecantile

tms = morecantile.tms.get("WebMercatorQuad")
morecantile_bounds = list(tms.xy_bounds(10, 9, 5))
print("morecantile boumds: ", morecantile_bounds)


src_path = "https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256pxWEBP.tif"

with rasterio.open(src_path) as src:
    col_off, row_off, w, h = windows.from_bounds(
        *morecantile_bounds, transform=src.transform, width=256, height=256,
    ).flatten()
    print("Window: ", col_off, row_off, w, h)
    
    w = windows.Window(
        col_off=round(col_off),
        row_off=round(row_off),
        width=round(w),
        height=round(h),
    )
    new_bounds = src.window_bounds(w)
    print("Corrected Window: ", *w.flatten())
    print("Corrected bounds: ", new_bounds)
    print("diff: ", [abs(new_bounds[ii] - morecantile_bounds[ii]) for ii in range(4)])

morecantile boumds:  [-7514065.6285459455, 7514065.6285459455, -6261721.35712162, 8766409.89997027]
Window:  1024.000000000005 4607.9999999999945 511.9999999999991 512.0
Corrected Window:  1024 4608 512 512
Corrected bounds:  (-7514065.628545957, 7514065.628545931, -6261721.357121631, 8766409.899970258)
diff:  [1.1175870895385742e-08, 1.4901161193847656e-08, 1.0244548320770264e-08, 1.30385160446167e-08]
```

### Discussion
- Is the implementation OK ? 
- Do we want this to be a default ?
- is the. `0.0001` threshold ok or should it be based on the dataset resolution ? 
- Should the threshold be applied on the window instead of bounds ? 

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -18,7 +18,13 @@ from . import constants
 from .errors import AlphaBandWarning, PointOutsideBounds, TileOutsideBounds
 from .utils import _requested_tile_aligned_with_internal_tile as is_aligned
 from .utils import _stats as raster_stats
-from .utils import get_vrt_transform, has_alpha_band, has_mask_band, non_alpha_indexes
+from .utils import (
+    get_aligned_bounds,
+    get_vrt_transform,
+    has_alpha_band,
+    has_mask_band,
+    non_alpha_indexes,
+)
 
 
 def read(
@@ -174,6 +180,8 @@ def part(
             raise TileOutsideBounds(
                 "Dataset covers less than {:.0f}% of tile".format(cover_ratio * 100)
             )
+
+    bounds = get_aligned_bounds(src_dst, bounds, height, width, bounds_crs=dst_crs)
 
     vrt_transform, vrt_width, vrt_height = get_vrt_transform(
         src_dst, bounds, height, width, dst_crs=dst_crs
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -18,13 +18,7 @@ from . import constants
 from .errors import AlphaBandWarning, PointOutsideBounds, TileOutsideBounds
 from .utils import _requested_tile_aligned_with_internal_tile as is_aligned
 from .utils import _stats as raster_stats
-from .utils import (
-    get_aligned_bounds,
-    get_vrt_transform,
-    has_alpha_band,
-    has_mask_band,
-    non_alpha_indexes,
-)
+from .utils import get_vrt_transform, has_alpha_band, has_mask_band, non_alpha_indexes
 
 
 def read(
@@ -181,8 +175,6 @@ def part(
                 "Dataset covers less than {:.0f}% of tile".format(cover_ratio * 100)
             )
 
-    bounds = get_aligned_bounds(src_dst, bounds, height, width, bounds_crs=dst_crs)
-
     vrt_transform, vrt_width, vrt_height = get_vrt_transform(
         src_dst, bounds, height, width, dst_crs=dst_crs
     )
@@ -201,7 +193,7 @@ def part(
 
     out_height = height or vrt_height
     out_width = width or vrt_width
-    if padding > 0 and not is_aligned(src_dst, bounds, out_height, out_width):
+    if padding > 0 and not is_aligned(src_dst, bounds, out_height, out_width, dst_crs):
         vrt_transform = vrt_transform * Affine.translation(-padding, -padding)
         orig_vrt_height = vrt_height
         orig_vrt_width = vrt_width
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -245,6 +245,46 @@ def linear_rescale(
     return image * (omax - omin) + omin
 
 
+def get_aligned_bounds(
+    src_dst: Union[DatasetReader, DatasetWriter, WarpedVRT],
+    bounds: Tuple[float, float, float, float],
+    height: Optional[int] = None,
+    width: Optional[int] = None,
+    bounds_crs: CRS = WEB_MERCATOR_CRS,
+    threshold: float = 0.0001,
+):
+    """Correct bounds to align with internal block bounds."""
+    if not src_dst.is_tiled:
+        return bounds
+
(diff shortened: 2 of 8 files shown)


rio_tiler/reader.py
```diff
@@ -193,7 +193,7 @@ def part(
 
     out_height = height or vrt_height
     out_width = width or vrt_width
-    if padding > 0 and not is_aligned(src_dst, bounds, out_height, out_width):
+    if padding > 0 and not is_aligned(src_dst, bounds, out_height, out_width, dst_crs):
```
vincentsarago, line 196: small change to `_requested_tile_aligned_with_internal_tile` method to allow other CRS than WebMercator

rio_tiler/utils.py
```diff
@@ -161,6 +162,24 @@ def get_vrt_transform(
...
+            *bounds, transform=src_dst.transform, width=width, height=height,
+        ).flatten()
+
+        w = windows.Window(
+            round(col_off, window_precision),
+            round(row_off, window_precision),
+            round(w, window_precision),
+            round(h, window_precision),
+        )
+        bounds = src_dst.window_bounds(w)
```
vincentsarago, line 181: If dataset is tiled and the window defined by the bbox is aligned with the internal tile then we define update the bounds to match the one defined by the internal block. We do this by using the `round` of the window
```python
Window(
    col_off=8192.000000016182,
    row_off=12288.000000016851,
    width=2047.999999999971, 
    height=2047.999999999971
)
Window(col_off=8192.0, row_off=12288.0, width=2048.0, height=2048.0)
```

vincentsarago:
Last commit changed a bit the method and I think it's better this way (do everything in the get_vrt_transform method).

There is still some issue tho

in https://github.com/cogeotiff/rio-tiler/blob/3dd6fc8b5ae789b499c41fc4b7973f48b35cebbc/rio_tiler/utils.py#L191-L207

the `vrt_width` and `vrt_height` can still be bigger than what we need

## To Do 
- fix the width/height rounding issue
- write test

vincentsarago:
Last commit `fix` the vrt size rounding issue. I basically took the same approach as in Terracotta https://github.com/DHI-GRAS/terracotta/blob/018c4f7a059086e0eee6f923370b7a03cd8ff70a/terracotta/drivers/raster_base.py#L475-L476

```
$ tilebench profile https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256pxWEBP.tif --tile 5-10-9                   
{"LIST": {"count": 0}, "HEAD": {"count": 1}, "GET": {"count": 2, "bytes": 32768, "ranges": ["0-16383", "229376-245759"]}, "Timing": 0.45401811599731445}

$ tilebench profile https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256px_COG.tif  --tile 5-10-9                   
{"LIST": {"count": 0}, "HEAD": {"count": 1}, "GET": {"count": 3, "bytes": 65536, "ranges": ["0-16383", "16384-49151", "376832-393215"]}, "Timing": 0.5621631145477295}
```

ref: https://github.com/cogeotiff/rio-tiler/issues/164

The latest change results in output size changes so the CI is failing.

vincentsarago:
Alright, I've been using this in production for couple days now. I didn't see any problem so I'm going to go ahead and merge this.

PR #357, opened 2021-03-03
added cog type from pystac

emmanuelmathot:
**Proposed Changes:**

***Adding pystac COG type***
in [pystac](https://github.com/stac-utils/pystac/blob/develop/pystac/media_type.py#L10) COG is defined as `image/tiff; profile=cloud-optimized; application=geotiff` and thus not recognized in rio-tiler STAC reader. the COG type defined in pystac 0.5.4 is then added for compatibility.

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## 2.0.4 (?)
+
+* Added [pystac.MediaType.COG](https://github.com/stac-utils/pystac/blob/master/pystac/media_type.py#L4) in supported types by STAC reader
+
 ## 2.0.3 (2021-02-19)
 
 * Reduce the number of `.read()` calls for dataset without nodata value (https://github.com/cogeotiff/rio-tiler/pull/355)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,6 @@
 * fix bad type definition in `rio_tiler.colormap.ColorMaps` data (https://github.com/cogeotiff/rio-tiler/issues/359)
 * add `rio_tiler.colormap.parse_color` function to parse HEX color (https://github.com/cogeotiff/rio-tiler/issues/361)
 
-
 ## 2.0.3 (2021-02-19)
 
 * Reduce the number of `.read()` calls for dataset without nodata value (https://github.com/cogeotiff/rio-tiler/pull/355)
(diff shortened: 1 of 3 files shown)


emmanuelmathot:
The best way should be to analyze properly the content-type matching the media-type ans all the parameters (profile and application). Here I added the possible types encountered and the one defined in pystac in case it evolves in the future.

geospatial-jeff (approved):

vincentsarago:
@emmanuelmathot thanks for the PR, do we have to add more tests?

kylebarron (approved):

rio_tiler/io/stac.py
```diff
@@ -19,6 +19,8 @@
 DEFAULT_VALID_TYPE = {
     "image/tiff; application=geotiff",
     "image/tiff; application=geotiff; profile=cloud-optimized",
+    "image/tiff; profile=cloud-optimized; application=geotiff",
+    pystac.MediaType.COG,
```
kylebarron, line 23: Wait why are you adding both here? Aren't they identical?

emmanuelmathot: yes, they are but this ensures the compatibility even it changes in a future version of pystac.

vincentsarago, line 23: Would it be cleaner to do `pystac.MediaType.COG.value` ?

vincentsarago:
@emmanuelmathot is this ready? I'm planning a release today!

CHANGES.md
```diff
@@ -1,8 +1,10 @@
-## 2.0.4 (TBD)
+## 2.0.4 (2021-03-09)
 
+* Added [pystac.MediaType.COG](https://github.com/stac-utils/pystac/blob/master/pystac/media_type.py#L4) in supported types by STAC reader
 * fix bad type definition in `rio_tiler.colormap.ColorMaps` data (https://github.com/cogeotiff/rio-tiler/issues/359)
 * add `rio_tiler.colormap.parse_color` function to parse HEX color (https://github.com/cogeotiff/rio-tiler/issues/361)
 
+
```
vincentsarago, line 7:
```suggestion
```

rio_tiler/io/stac.py
vincentsarago, line 23:
```suggestion
     pystac.MediaType.COG.value,
```

PR #361, opened 2021-03-05
add parse_color function

vincentsarago:
ref https://github.com/developmentseed/titiler/pull/262

This PR adds a function to parse colormap value from HEX for list.


geospatial-jeff (approved):

rio_tiler/colormap.py
```diff
@@ -133,6 +139,67 @@ def apply_discrete_cmap(
     return data[:-1], data[-1]
 
 
+def parse_color(value: Union[List[int], str]) -> List[int]:
```
kylebarron, line 142: Should the type be a tuple because input is always length 3 or 4 and output is always length 4?

rio_tiler/colormap.py
```diff
@@ -133,6 +139,67 @@ def apply_discrete_cmap(
...
+
+        >>> parse_color("#FF0000")
+        [255, 0, 0, 255]
+
+        >>> parse_color([255, 255, 255])
+        [255, 255, 255, 255]
+
+    """
+    if isinstance(value, str):
+        if re.match("^#[a-fA-F0-9]{3,4}$", value):
```
kylebarron, line 166: IMO a regex is overkill here. You can just assert that the string starts with `#` (if you want) and then if the string has <=4 more characters, you treat it as the first type, otherwise the second. And if an error is created that gets propagated to the user. But not a strong preference.

kylebarron (approved):

rio_tiler/colormap.py
vincentsarago: FYI, that's how matplotlib does: https://github.com/matplotlib/matplotlib/blob/fa73bc09d62aef99842bbbd13dca687ac5bbd0b9/lib/matplotlib/colors.py#L237-L265

I think overkill regex are fine :-)

rio_tiler/colormap.py
```diff
@@ -133,6 +139,68 @@ def apply_discrete_cmap(
...
+        ]
+
+    if len(rgba) > 4 or len(rgba) < 3:
+        raise InvalidColorFormat(f"Invalid color format: {rgba}")
+
+    rgba = tuple(rgba)
+    if len(rgba) == 3:
+        rgba += (255,)
+
+    return rgba  # type: ignore
```
vincentsarago, line 201: I couldn't keep mypy happy here because it lost the track on the number of items in rgba 🤷‍♂️

geospatial-jeff (approved):

PR #377, opened 2021-04-01
add python3.9 in CI

vincentsarago:


kylebarron (approved):

vincentsarago:
Tests are failing in 3.9 because of the change from #370 

> resource is the name of the resource to open within package; it may not contain path separators and it may not have sub-resources (i.e. it cannot be a directory).

I guess it work in 3.8 ... but it shouldn't have

PR #378, opened 2021-04-01
Fix colormap loading

kylebarron:
Fixes colormap loading on 3.9


vincentsarago:
went another for another way #379

Issue #348, opened 2021-02-11
read low resolution version of the data


vincentsarago:
By design rio-tiler assume you are working within the native zoom range of your data (raw data = Max res, max res/number of overviews ~= minzoom). 

When reading a file, we always use a VRT to align the data to the `bounds`/`size` we want to read. This is all defined in https://github.com/cogeotiff/rio-tiler/blob/ccf6aa002fc0c3525c9889cffbb9ca26689ba0b9/rio_tiler/utils.py#L172-L188

(this part was copie from https://github.com/mapbox/rasterio/issues/1373#issuecomment-398261502) 

if you try to read a really low version of your data, we will first create a huge VRT and then ask to read a decimated part of it (this is to make sure we read the overview). 

To be honest, for now I'm not sure there is a solution for this.

PR #381, opened 2021-04-07
add warning when trying to create huge VRT

vincentsarago:
ref #348 

Still not sure how to resolve this issue, but for now here I added a warning when the output resolution is bigger than the small overview resolution. 

The main problem is that in this case, rio-tiler will create a Huge VRT (matching the input bounds) and use `out_shape` https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/reader.py#L101 to fetch the data for the right overview. But when the output resolution is > to the min resolution for the COG, we will create a bigger VRT and still read part like 256x256 while in reality with might need only 10x10 pixel resampled to 256x256.


rio_tiler/utils.py
```diff
@@ -206,6 +207,15 @@ def get_vrt_transform(
     vrt_height = max(1, round((s - n) / h_res))
     vrt_transform = from_bounds(w, s, e, n, vrt_width, vrt_height)
 
+    # Warns when the read resolution (tile, part) is lower than the
+    # resolution of the smaller overview. This will result in rasterio/rio-tiler
+    # creating a big WarpedVRT and trying to read a small part of it.
```
kylebarron, line 212: I'd put this in the second line of the warning message as well. Just saying

> Trying to fetch lower resolution than min overview resolution

doesn't really tell anything to the user, and as a user I'd be confused why I'm getting warned at.

kylebarron (approved):

rio_tiler/utils.py
vincentsarago: yeah, TBH I'm still not sure about the warning, I hope I'll be able to find a solution instead 😄

PR #386, opened 2021-04-16
add warning when no materialized overview are found

vincentsarago:


rio_tiler/io/cogeo.py
```diff
@@ -109,6 +109,14 @@ def __attrs_post_init__(self):
         if self.colormap is None:
             self._get_colormap()
 
+        if min(
+            self.dataset.width, self.dataset.height
+        ) > 512 and not self.dataset.overviews(1):
+            warnings.warn(
+                "The dataset has no materialized Overviews. rio-tiler performances might be impacted",
```
vincentsarago, line 116:
```suggestion
                "The dataset has no Overviews. rio-tiler performances might be impacted",
```

I added `materialized` because in case of VRT, the COG can have overviews but they are not reported at the VRT level. I think removing this word is fine. I'll add a not in the documentation about rio-tiler support for VRT 

cc @geospatial-jeff

PR #387, opened 2021-04-17
add more info in COGReader.info()

### LINKED ISSUE #384: Add width, height, count in `Info` ?
https://github.com/cogeotiff/rio-tiler/blob/56f50abdd06a4f03a18ee1e63c0db233c474fffe/rio_tiler/models.py#L53-L63

vincentsarago:
closes #384 

This PR does:
- change Info model config to allow `additional` informations
- add width, height, count, overviews and nodata_value in COGReader.info()

### DIFF
--- a/rio_tiler/models.py
+++ b/rio_tiler/models.py
@@ -65,7 +65,7 @@ class Info(SpatialInfo):
     class Config:
         """Config for model."""
 
-        extra = "ignore"
+        extra = "allow"
         use_enum_values = True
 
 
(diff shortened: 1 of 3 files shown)


rio_tiler/models.py
```diff
@@ -65,7 +65,7 @@ class Info(SpatialInfo):
     class Config:
         """Config for model."""
 
-        extra = "ignore"
+        extra = "allow"
```
vincentsarago, line 68: width, height, overviews are not set in the model because for `mosaic` or `stac` it doesn't really make sense. It's easier to just `allow` extra value in the model

kylebarron (approved):

PR #391, opened 2021-05-13
add autorescaling to avoid image creation errors

### LINKED ISSUE #390: rescale data to avoid Driver error
```python
In [38]: img = ImageData(numpy.zeros((1,256,256), dtype="int8")).render("jpeg")

...rio-tiler/rio_tiler/utils.py in render(data, mask, img_format, colormap, **creation_options)
    378     with MemoryFile() as memfile:
    379         with memfile.open(**output_profile) as dst:
--> 380             dst.write(data, indexes=list(range(1, count + 1)))
    381             # Use Mask as an alpha band
    382             if mask is not None:

...rasterio/rasterio/_io.pyx in rasterio._io.DatasetWriterBase.write()

ValueError: the array's dtype 'int8' does not match the file's dtype 'uint8'
```

https://github.com/cogeotiff/rio-tiler/blob/6c8da5e88ca562205998fae73aa42a0e9a0014b5/rio_tiler/utils.py#L383

vincentsarago:
closes #390 

This PR does:
- [x] add data type rescaling when we know the output driver will raise an error (e.g float -> uint8 for PNG)

**Why the `auto-rescaling` is only in `ImageData.render` and not in `rio_tiler.utils.render`?**

The rescale method (`rio_tiler.utils.linear_rescale`) requires a `mask` in argument but in `rio_tiler.utils.render` the mask is either optional or deleted. In the ImageData object, the mask will always be defined.

### DIFF
--- a/rio_tiler/models.py
+++ b/rio_tiler/models.py
@@ -18,7 +18,7 @@ from rio_color.utils import scale_dtype, to_math_type
 
 from .constants import ColorTuple, NumType
 from .errors import InvalidDatatypeWarning
-from .utils import _chunks, linear_rescale, render
+from .utils import linear_rescale, render
 
 
 class NodataTypes(str, Enum):
@@ -194,26 +194,24 @@ class ImageData:
         self,
         data: numpy.ndarray,
         mask: numpy.ndarray,
-        in_range: Tuple[NumType, NumType],
-        out_range: Tuple[NumType, NumType] = (0, 255,),
+        in_range: Sequence[Tuple[NumType, NumType]],
+        out_range: Sequence[Tuple[NumType, NumType]] = ((0, 255),),
         out_dtype: Union[str, numpy.number] = "uint8",
     ):
         """Rescale data."""
         nbands = data.shape[0]
 
-        rescale_arr = tuple(_chunks(in_range, 2))
-        if len(rescale_arr) != nbands:
-            rescale_arr = ((rescale_arr[0]),) * nbands
+        if len(in_range) != nbands:
+            in_range = ((in_range[0]),) * nbands
 
-        out_rescale_arr = tuple(_chunks(out_range, 2))
-        if len(out_rescale_arr) != nbands:
-            out_rescale_arr = ((out_rescale_arr[0]),) * nbands
+        if len(out_range) != nbands:
+            out_range = ((out_range[0]),) * nbands
 
         for bdx in range(nbands):
             data[bdx] = numpy.where(
                 mask,
                 linear_rescale(
-                    data[bdx], in_range=rescale_arr[bdx], out_range=out_rescale_arr[bdx]
+                    data[bdx], in_range=in_range[bdx], out_range=out_range[bdx]
                 ),
                 0,
             )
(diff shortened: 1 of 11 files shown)


rio_tiler/models.py
```diff
@@ -187,6 +190,36 @@ def transform(self):
...
+        data: numpy.ndarray,
+        mask: numpy.ndarray,
+        in_range: Tuple[NumType, NumType],
+        out_range: Tuple[NumType, NumType] = (0, 255,),
+        out_dtype: Union[str, numpy.number] = "uint8",
+    ):
+        """Rescale data."""
+        nbands = data.shape[0]
+
+        rescale_arr = tuple(_chunks(in_range, 2))
```
geospatial-jeff, line 204: Are `in_range` and `out_range` typed correctly in the function signature?  The call to `_chunks` indicates to me these parameters can have more than 2 elements (one set of rescale values per band).

vincentsarago: so basically if the input is `in_range=(1,255,1,255,1,255)`  the `chunks` method will return `((1, 255), (1,255), (1,255))`. I guess the type `Tuple[NumType, NumType]` is wrong because in fact there can be more than 2 elements 🤔

vincentsarago: maybe switching to `Sequence[NumType]` would be better ?

vincentsarago: OR, maybe we let this to the user, meaning the `in_range` should always be of type
`Sequence[Tuple[NumType]]` (e.g: `((0, 255),)` ), we will still match the number of bands if only one tuple is provided. 

This will be a breaking change, but I'm fine with it because we move to a `minor` version

geospatial-jeff: You can do `Tuple[NumType, ...]` to indicate a variable length tuple

vincentsarago: 👍 
I mean the `chunk` part is still weird and I feel it should go! 

IMO, input should be in form of `Sequence[Tuple[NumType, NumType]]`, Sequence should match the number of band or be 1.

geospatial-jeff: I agree with that, `Sequence[Tuple[NumType, NumType]]` is more explicit, as long as you are ok with the breaking change.

vincentsarago: @kylebarron @geospatial-jeff if we move to `Sequence[Tuple[NumType, NumType]]` do we need to validate the input of we let `type` checking do the trick?

geospatial-jeff: I don't think we need to validate, its clear from the type hint

rio_tiler/models.py
```diff
@@ -91,6 +94,20 @@ def to_coordsbbox(bbox) -> Optional[BoundingBox]:
     return BoundingBox(*bbox) if bbox else None
 
 
+# def validate_ranges(
```
vincentsarago, line 97: 😂

geospatial-jeff (approved):

geospatial-jeff (approved):

PR #403, opened 2021-07-07
re-use class methods to reduce code duplication

vincentsarago:
This PR does:
- simplify `COGReader.tile` and `COGReader.feature` to use `COGReader.part` method.

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -270,30 +270,16 @@ class COGReader(BaseReader):
             tilesize (int, optional): Output image size. Defaults to `256`.
             indexes (int or sequence of int, optional): Band indexes.
             expression (str, optional): rio-tiler expression (e.g. b1/b2+b3).
-            kwargs (optional): Options to forward to the `rio_tiler.reader.part` function.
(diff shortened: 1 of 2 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -477,10 +398,65 @@ def point(
         if expression:
             blocks = expression.lower().split(",")
             bands = [f"b{bidx}" for bidx in indexes]
-            point = apply_expression(blocks, bands, point).tolist()
+            point = apply_expression(blocks, bands, numpy.array(point)).tolist()
 
         return point
 
+    def tile(
```
kylebarron, line 405: Why did you reorder this method? Seems substantially similar to the previous `COGReader.tile` method, but since it's moved there's no easy way to see the actual diff.

kylebarron (approved):

rio_tiler/io/cogeo.py
vincentsarago: let me undo this

kylebarron (approved):

geospatial-jeff (approved):

PR #404, opened 2021-07-15
pass options to the fetch function

### LINKED ISSUE #397: Resource Authorization manager
## Context
In titiler, we often need to access protected resources (basic auth, JWT, cookies...). In our current setup, we use a mix of environement variables such as `GDAL_HTTP_AUTH`, `GDAL_HTTP_USERPWD` for cog I/O and netrc files for python requests (stac I/O). This is not very easy to maintain and we are currently limited to HTTP basic authentication with a unique user.

## Issue Description
Design and implement a possible homogeneous mechanism to authenticate requests to protected resources
Currently identified remote resource access:
### COG servers via gdal /vsicurl/ driver
GDAL /vsicurl/ configuration is not super flexible for setting dynamically HTTP headers but it seems to have some options such as GDAL_HTTP_HEADER_FILE

### STAC catalogs using python requests lib
This one is easier since the requests lib allows all HTTP settings

Maybe there is already a mechanism I missed. Let me know

vincentsarago:
closes #397 

@emmanuelmathot can you check this :-)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 * add support for setting the S3 endpoint url via the `AWS_S3_ENDPOINT` environment variables in `aws_get_object` function using boto3 (https://github.com/cogeotiff/rio-tiler/pull/394)
 * make `ImageStatistics.valid_percent` a value between 0 and 100 (instead of 0 and 1) (author @param-thakker, https://github.com/cogeotiff/rio-tiler/pull/400)
-* add `client_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
+* add `fetch_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
 
     ```python
     with STACReader("s3://...", fetch_options={"request_pays": True}):
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,13 @@
 ## Unreleased
 
 * add support for setting the S3 endpoint url via the `AWS_S3_ENDPOINT` environment variables in `aws_get_object` function using boto3 (https://github.com/cogeotiff/rio-tiler/pull/394)
-
 * make `ImageStatistics.valid_percent` a value between 0 and 100 (instead of 0 and 1) (author @param-thakker, https://github.com/cogeotiff/rio-tiler/pull/400)
+* add `client_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
+
+  ```python
+  with STACReader("s3://...", fetch_options={"request_pays": True}):
+    pass
+  ```
 
 ## 2.1.0 (2021-05-17)
 
--- a/rio_tiler/io/stac.py
+++ b/rio_tiler/io/stac.py
@@ -38,7 +38,7 @@ def fetch(filepath: str, **kwargs: Any) -> Dict:
 
     Args:
         filepath (str): STAC item URL.
-        kargs (any): additional options to pass to client.
+        kwargs (any): additional options to pass to client.
 
     Returns:
         dict: STAC Item content.
--- a/rio_tiler/io/stac.py
+++ b/rio_tiler/io/stac.py
@@ -132,7 +132,7 @@ class STACReader(MultiBaseReader):
         exclude_asset_types (set of string, optional): Exclude some assets base on their type.
         reader (rio_tiler.io.BaseReader, optional): rio-tiler Reader. Defaults to `rio_tiler.io.COGReader`.
         reader_options (dict, optional): Additional option to forward to the Reader. Defaults to `{}`.
-        fetch_options (dict, optional): Options to pass to the client (requests, boto3 ...) fetching the STAC Items. Defaults to `{}`.
+        fetch_options (dict, optional): Options to pass to `rio_tiler.io.stac.fetch` function fetching the STAC Items. Defaults to `{}`.
 
     Examples:
         >>> with STACReader(stac_path) as stac:
--- a/rio_tiler/io/stac.py
+++ b/rio_tiler/io/stac.py
@@ -2,7 +2,8 @@
 
 import functools
 import json
-from typing import Dict, Iterator, Optional, Set, Type, Union
+import os
+from typing import Any, Dict, Iterator, Optional, Set, Type, Union
 from urllib.parse import urlparse
 
 import attr
@@ -31,13 +32,14 @@ DEFAULT_VALID_TYPE = {
 
 
 @functools.lru_cache(maxsize=512)
-def fetch(filepath: str) -> Dict:
+def fetch(filepath: str, **kwargs: Any) -> Dict:
     """Fetch STAC items.
 
     A LRU cache is set on top of this function.
 
     Args:
         filepath (str): STAC item URL.
+        kargs (any): additional options to pass to client.
 
     Returns:
         dict: STAC Item content.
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -1,8 +1,13 @@
...
 
 * add support for setting the S3 endpoint url via the `AWS_S3_ENDPOINT` environment variables in `aws_get_object` function using boto3 (https://github.com/cogeotiff/rio-tiler/pull/394)
-
 * make `ImageStatistics.valid_percent` a value between 0 and 100 (instead of 0 and 1) (author @param-thakker, https://github.com/cogeotiff/rio-tiler/pull/400)
+* add `client_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
+
+  ```python
+  with STACReader("s3://...", fetch_options={"request_pays": True}):
+    pass
+  ```
```
kylebarron, line 10: Four space indent?
```suggestion
    ```python
    with STACReader("s3://...", fetch_options={"request_pays": True}):
        pass
    ```
```

rio_tiler/io/stac.py
```diff
@@ -31,13 +32,14 @@
...
 @functools.lru_cache(maxsize=512)
-def fetch(filepath: str) -> Dict:
+def fetch(filepath: str, **kwargs: Any) -> Dict:
     """Fetch STAC items.
 
     A LRU cache is set on top of this function.
 
     Args:
         filepath (str): STAC item URL.
+        kargs (any): additional options to pass to client.
```
kylebarron, line 42:
```suggestion
        kwargs (any): additional options to pass to client.
```

rio_tiler/io/stac.py
```diff
@@ -47,10 +49,17 @@ def fetch(filepath: str) -> Dict:
     if parsed.scheme == "s3":
         bucket = parsed.netloc
         key = parsed.path.strip("/")
-        return json.loads(aws_get_object(bucket, key))
+
+        client = kwargs.get("client")
+        request_pays = kwargs.get(
+            "request_pays", os.environ.get("AWS_REQUEST_PAYER", False),
+        )
```
kylebarron, line 56: Why should the kwarg be `request_pays` instead of `request_payer`? Seems better to match the AWS key?

kylebarron: Or I guess AWS is inconsistent between their shell env and boto3 API's?

kylebarron: Wait, `request_pays` is just your wording in https://github.com/cogeotiff/rio-tiler/blob/3db728210994c5ddc5cd22d117626e9ef2e9ba4d/rio_tiler/utils.py#L34? If so it seems better for the public API to be consistent with AWS APIs, in which case we should use `request_payer`?

rio_tiler/io/stac.py
```diff
@@ -130,7 +139,8 @@ class STACReader(MultiBaseReader):
         include_asset_types (set of string, optional): Only include some assets base on their type.
         exclude_asset_types (set of string, optional): Exclude some assets base on their type.
         reader (rio_tiler.io.BaseReader, optional): rio-tiler Reader. Defaults to `rio_tiler.io.COGReader`.
-        reader_options (dict, optional): additional option to forward to the Reader. Defaults to `{}`.
+        reader_options (dict, optional): Additional option to forward to the Reader. Defaults to `{}`.
+        fetch_options (dict, optional): Options to pass to the client (requests, boto3 ...) fetching the STAC Items. Defaults to `{}`.
```
kylebarron, line 143: I think somewhere you should document the special behavior for `client` and `request_pays` kwargs for AWS. Something like
"When using a STAC object with HTTPS urls, kwargs is passed to `requests.get` after the url. When using an object with S3 urls, `client` and `request_pays` kwargs are passed to boto3's `get_object`"

kylebarron (approved):

rio_tiler/io/stac.py
vincentsarago: I don't think it's good to change this now!

geospatial-jeff (approved):

geospatial-jeff (approved):

geospatial-jeff:
Not a strong opinion, but it feels weird to me that `fetch_options` wraps both `requests` and `boto3` kwargs but only specific keys from `fetch_options` are forwarded to boto3 while all keys go to requests.  I think this could cause some confusion:

A user might pass a boto3 session kwarg and expect it be passed to client but it won't (assuming they aren't passing the client directly).  If its really just s3 requester pays that we want to expose I think it makes more sense to have a single boolean argument (`s3_requester_pays`) and leave `fetch_options` just for requests (and maybe rename it to `requests_http_kwargs` or something more explicit)

vincentsarago:
Ok let me try something else!

CHANGES.md
```diff
@@ -1,8 +1,13 @@
 ## Unreleased
 
 * add support for setting the S3 endpoint url via the `AWS_S3_ENDPOINT` environment variables in `aws_get_object` function using boto3 (https://github.com/cogeotiff/rio-tiler/pull/394)
-
 * make `ImageStatistics.valid_percent` a value between 0 and 100 (instead of 0 and 1) (author @param-thakker, https://github.com/cogeotiff/rio-tiler/pull/400)
+* add `client_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
```
vincentsarago, line 5:
```suggestion
* add `fetch_options` to `STACReader` to allow custom configuration to the fetch client (https://github.com/cogeotiff/rio-tiler/pull/404)
```

geospatial-jeff (approved):

rio_tiler/io/stac.py
```diff
@@ -47,10 +48,10 @@ def fetch(filepath: str) -> Dict:
     if parsed.scheme == "s3":
         bucket = parsed.netloc
         key = parsed.path.strip("/")
-        return json.loads(aws_get_object(bucket, key))
+        return json.loads(aws_get_object(bucket, key, **kwargs))
 
     elif parsed.scheme in ["https", "http", "ftp"]:
-        return requests.get(filepath).json()
+        return requests.get(filepath, **kwargs).json()
```
kylebarron, line 54: Seems like a possible issue with this is that a user can't use a single `fetch_options` for data that _sometimes_ has HTTPS urls and _sometimes_ has `s3:` urls... 

I.e. if you have a `params` key in your `fetch_options` but your data unexpectedly has an asset with an `s3://` url, then `params` will be passed to `aws_get_object` and fail, right?

vincentsarago: fetch_options is not used at the asset level but just to get the `Items`

kylebarron: A solution we've used for problems like this in loaders.gl is to have nested options keys, so for example `fetch_options` would take options that look like
```js
{
	'https': {
		params: {}
	},
	's3': {
		request_pays: True,
	}
}
```
where the `https` key gets passed to `requests` and the `s3` key gets passed to `aws_get_object`.

kylebarron: > fetch_options is not used at the asset level but just to get the `Items`

oh, nevermind then. If it's just to get the STAC Item itself, it shouldn't be too hard for the end user to know where their stac data is coming from

kylebarron (approved):

rio_tiler/io/stac.py
vincentsarago:
```suggestion
        fetch_options (dict, optional): Options to pass to `rio_tiler.io.stac.fetch` function fetching the STAC Items. Defaults to `{}`.
```

emmanuelmathot:
Sorry for the missed review. Thank you for the effort!

PR #405, opened 2021-07-19
Allow float tile_buffer

bstadlbauer:
Some tile data such a certain digital elevation models (DEMs) default to an uneven tile size (e.g. 257x257). They have an overlap of one pixel on each side, and assume that each pixel is not an "area" but a point measurement. It would be nice, if `COGReader.tile()` would support this use case.

I found the `tile_buffer` parameter, which would do this exact thing. However, as it only supports integers, only overlaps of two pixel are possible. This PR changes the signature of `tile_buffer` to be `Optional[Union[int, float]]` where the `float` case is restricted to halves of integers, to ensure an integer `tilesize`


vincentsarago:
Thanks @bstadlbauer 🙏 

I wonder if we could just allow integer and explicitly tell that the value will be `divided by 2` 🤷‍♂️ 

Adding a test to check if the tile_buffer is a multiple of 0.5 seems a bit weird TBH. 

I'll wait for @kylebarron and @geospatial-jeff input

bstadlbauer:
That's also something that I thought about. For me, both versions don't seem quite right so I went with the one not breaking the existing API. 

I am flexible here, just let me know and I adjust the code as required.

TheCthulhuKid (approved):

PR #407, opened 2021-07-29
Fix alpha band values when storing `Uint16` data in PNG

### LINKED ISSUE #406: [Render] Wrong values for PNG alpha band
In rio-tiler/gdal/rasterio we assume alpha band to be between 0 and 255. But in the PNG format the values is between 0 and  `2sampledepth - 1` (255 or 65535).

> An alpha channel included in the image array has 8-bit or 16-bit samples, the same size as the other samples. The alpha sample for each pixel is stored immediately following the greyscale or RGB samples of the pixel. An alpha value of zero represents full transparency, and a value of 2sampledepth - 1 represents full opacity. Intermediate values indicate partially transparent pixels that can be composited against a background image to yield the delivered image.

ref https://www.w3.org/TR/PNG/#6AlphaRepresentation

This is important, especially when we export `Uin16` data to PNG, and right now, it will result in transparent PNG (because 255 is close to 0 while it should be 65535

vincentsarago:
closes #406

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -344,6 +344,9 @@ def render(
     if img_format == "WEBP" and data.shape[0] == 1:
         data = numpy.repeat(data, 3, axis=0)
 
+    if img_format == "PNG" and data.dtype == "uint16" and mask is not None:
+        mask = linear_rescale(mask, (0, 255), (0, 65535)).astype("uint16")
+
     elif img_format == "JPEG":
         mask = None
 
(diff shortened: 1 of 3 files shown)


rio_tiler/utils.py
```diff
@@ -344,6 +344,9 @@ def render(
     if img_format == "WEBP" and data.shape[0] == 1:
         data = numpy.repeat(data, 3, axis=0)
 
+    if img_format == "PNG" and data.dtype == "uint16" and mask is not None:
+        mask = linear_rescale(mask, (0, 255), (0, 65535)).astype("uint16")
```
vincentsarago, line 348: To ensure we keep values for 0 to 255 we apply a linear rescaling to 0 -> 65535

kylebarron (approved):

PR #410, opened 2021-08-13
Bring mypy out of pre-commit to type check against imports

kylebarron:
Testing on CI; I can't figure out how to add a section to `tox.ini`. I added (locally, didn't commit)
```
[testenv:typing]
extras = test
commands=
    python -m mypy --config-file setup.cfg -p rio_tiler
deps=
    numpy
```
but it didn't seem to get run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       # Run mypy using the version of Python in `PATH`
       - name: Run mypy
         run: |
-            python -m pip install '.[test]'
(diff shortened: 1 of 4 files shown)


kylebarron:
Interestingly.. there are no numpy type errors with numpy 1.19; a few on 1.20; and many on 1.21. Tox [is installing](https://github.com/cogeotiff/rio-tiler/runs/3324405145#step:5:23) numpy 1.19.5, but I can't figure out why. There's no restriction on numpy, so I'd expect it to install 1.21 (and that's what it does for me locally).

kylebarron:
I tried to pin to numpy 1.21 in tests only... and tox can't understand that apparently 🤷‍♂️ (though pip can!).

kylebarron:
```
ERROR: Could not find a version that satisfies the requirement numpy<1.22,>=1.21
ERROR: No matching distribution found for numpy<1.22,>=1.21
```
zero clue why pip can't find any numpy version past 1.19.5

kylebarron:
> zero clue why pip can't find any numpy version past 1.19.5

This is because 1.19.5 was the last one to support Python 3.6.

🤦‍♂️

vincentsarago:
done in #461

PR #411, opened 2021-08-13
Start fixing types, pt 1

kylebarron:


vincentsarago:
made the changes in #460

PR #413, opened 2021-08-18
Rename COGReader's input parameter in example

plant99:
Change COGReader's input parameter from `my.tif` to `file` as passed in parameters.


vincentsarago (approved):
🤦

PR #422, opened 2021-09-14
No max size

vincentsarago:
ref https://github.com/cogeotiff/rio-tiler/projects/3#card-68449203, https://github.com/cogeotiff/rio-tiler/projects/3#card-68449132

This PR removes `max_size` defaults (1024) to read full resolution data by default. It also reorder/add options in COGReader method for documentation.

### DIFF
--- a/rio_tiler/colormap.py
+++ b/rio_tiler/colormap.py
@@ -25,7 +25,7 @@ except ImportError:
 EMPTY_COLORMAP: Dict = {i: [0, 0, 0, 0] for i in range(256)}
 
 DEFAULT_CMAPS_FILES = {
-    f.stem: str(f) for f in (resources_files(__package__) / "cmap_data").glob("*.npy")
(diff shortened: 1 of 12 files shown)


geospatial-jeff (approved):

kylebarron (approved):

rio_tiler/colormap.py
```diff
@@ -25,7 +25,7 @@
 EMPTY_COLORMAP: Dict = {i: [0, 0, 0, 0] for i in range(256)}
 
 DEFAULT_CMAPS_FILES = {
-    f.stem: str(f) for f in (resources_files(__package__) / "cmap_data").glob("*.npy")
+    f.stem: str(f) for f in (resources_files(__package__) / "cmap_data").glob("*.npy")  # type: ignore
```
vincentsarago, line 28: seen this also in timvt https://github.com/developmentseed/timvt/pull/58/commits/aab6c5b0d7d779e808009ac7aac10d0b4f60abbe
For now we can ignore but it would be good to revisit to see if there is a better way

PR #418, opened 2021-09-09
switch to morecantile3

vincentsarago:
ref #417

PR #423, opened 2021-09-14
make sure we return uint8 when applying colormap

### LINKED ISSUE #421: Output array after applying colormap should be of uint8 type
https://github.com/cogeotiff/rio-tiler/blob/2a903d8aecb0a6093005e219f6bee9dce96b5b24/rio_tiler/colormap.py#L81-L112

vincentsarago:
closes #421


rio_tiler/colormap.py
```diff
@@ -141,7 +141,7 @@ def apply_discrete_cmap(
     for k, v in colormap.items():
         res[data[0] == k] = v
 
-    data = numpy.transpose(res, [2, 0, 1])
+    data = numpy.transpose(res, [2, 0, 1]).astype("uint8")
```
vincentsarago, line 144: we make sure the output is in `uint8`

rio_tiler/models.py
```diff
@@ -285,32 +292,35 @@ def render(self, add_mask: bool = True, img_format: str = "PNG", **kwargs) -> by
...
-            "uint8",
-            "int16",
-            "uint16",
-        ]:
-            warnings.warn(
-                f"Invalid type: `{data.dtype}` for the `{img_format}` driver. Data will be rescaled using min/max type bounds.",
-                InvalidDatatypeWarning,
-            )
-            data = self._rescale(data, self.mask, in_range=datatype_range)
+        if not colormap:
```
vincentsarago, line 295: if there is a colormap passed we don't need to auto-rescale

vincentsarago:
test are failing because of recent change in rasterio (see: https://github.com/mapbox/rasterio/issues/2289) but was fixed in  https://github.com/cogeotiff/rio-tiler/pull/418 by changing the dateline dataset 

let me see if I can bring the change we push in rio-tiler-v3 branch

Issue #424, opened 2021-09-20
Better statistics 


vincentsarago:
https://github.com/developmentseed/titiler/blob/master/src/titiler/core/titiler/core/utils.py#L61-L120

PR #425, opened 2021-09-20
Metadata deprecation

### LINKED ISSUE #420: depreciate the `BaseReader.metadata` method
The metadata method is the combination of `info` and `statistics`. It doesn't really provide any real advantage and kinda make evolution of the internal API more complexe (any change in the stats or info method have to be integrated in the metadata method)

vincentsarago:
closes #420


vincentsarago:
Once this one is merged I'll work on #424

kylebarron (approved):

PR #427, opened 2021-09-28
add new statistics methods and band names in ImageData object

vincentsarago:
This PR will bring some important change to the `statistics` method in rio-tiler

- replaces `stats` by `statistics` method (in baseclass)
- add new `BandStatistics` model (replace `ImageStatistics`)
- add `band_names` in ImageData class (statistics can be get for expression or indexes so we need a way to forward the band information from the ImageData object to the statistics method)
- `statistics` method is only able to return stats for the full dataset (no bounds nor feature)

## To Do

- [ ] validate ☝️ changes
- [ ] update base classes (deprecate `stats` and add `statistics`)
- [ ] update tests
- [ ] update documentation
- [ ] start docs for v2 to v3

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -20,8 +20,14 @@ from .. import reader
 from ..constants import WEB_MERCATOR_TMS, WGS84_CRS, BBox, Indexes, NoData
 from ..errors import ExpressionMixingWarning, NoOverviewWarning, TileOutsideBounds
 from ..expression import apply_expression, parse_expression
-from ..models import ImageData, ImageStatistics, Info
-from ..utils import create_cutline, has_alpha_band, has_mask_band
+from ..models import BandStatistics, ImageData, ImageStatistics, Info
+from ..utils import (
+    create_cutline,
+    get_array_statistics,
+    get_bands_names,
+    has_alpha_band,
+    has_mask_band,
+)
 from .base import BaseReader
 
 
@@ -236,6 +242,11 @@ class COGReader(BaseReader):
             rio_tiler.models.ImageStatistics: bands statistics.
 
         """
+        warnings.warn(
+            "`stats` method will be removed and replaced by `statistics` in rio-tiler v3.0.0",
+            DeprecationWarning,
+        )
+
         kwargs = {**self._kwargs, **kwargs}
 
         hist_options = hist_options or {}
@@ -249,6 +260,46 @@ class COGReader(BaseReader):
             self.dataset, percentiles=(pmin, pmax), hist_options=hist_options, **kwargs,
         )
         return {b: ImageStatistics(**s) for b, s in stats.items()}
+
+    def statistics(
+        self,
+        categorical: bool = False,
+        categories: Optional[List[float]] = None,
+        percentiles: List[int] = [2, 98],
+        hist_options: Optional[Dict] = None,
+        max_size: int = 1024,
+        **kwargs: Any,
+    ) -> Dict[str, BandStatistics]:
+        """Return bands statistics from a COG.
+
+        Args:
+
+
+            hist_options (dict, optional): Options to forward to numpy.histogram function.
+            kwargs (optional): Options to forward to `self.preview`.
+
+        Returns:
+            Dict[str, rio_tiler.models.BandStatistics]: bands statistics.
+
+        """
+        kwargs = {**self._kwargs, **kwargs}
+
+        data = self.preview(max_size=max_size, **kwargs)
+
+        hist_options = hist_options or {}
+
+        stats = get_array_statistics(
+            data.as_masked(),
+            categorical=categorical,
+            categories=categories,
+            percentiles=percentiles,
+            **hist_options,
+        )
+
+        return {
+            f"{data.band_names[ix]}": BandStatistics(**stats[ix])
+            for ix in range(len(stats))
+        }
 
     def tile(
         self,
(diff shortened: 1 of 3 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -250,6 +261,46 @@ def stats(
...
+            hist_options (dict, optional): Options to forward to numpy.histogram function.
+            kwargs (optional): Options to forward to `self.preview`.
+
+        Returns:
+            Dict[str, rio_tiler.models.BandStatistics]: bands statistics.
+
+        """
+        kwargs = {**self._kwargs, **kwargs}
+
+        data = self.preview(max_size=max_size, **kwargs)
```
vincentsarago, line 287: the `statistics` method will only get stats for the full dataset (using preview) but we can set `max_size` to control the resolution the user wants

rio_tiler/io/cogeo.py
```diff
@@ -250,6 +261,46 @@ def stats(
...
+        stats = get_array_statistics(
+            data.as_masked(),
+            categorical=categorical,
+            categories=categories,
+            percentiles=percentiles,
+            **hist_options,
+        )
+
+        return {
+            f"{data.band_names[ix]}": BandStatistics(**stats[ix])
```
vincentsarago, line 300: we use `band_names` to keep trace of the `bands` we return

rio_tiler/models.py
```diff
@@ -83,6 +83,30 @@ class ImageStatistics(RioTilerBaseModel):
...
+    unique: float
+    histogram: List[List[NumType]]
+    valid_percent: float
+    masked_pixels: float
+    valid_pixels: float
+
+    class Config:
+        """Config for model."""
+
+        extra = "allow"
```
vincentsarago, line 107: we allow extra for `percentile_{}` values

rio_tiler/utils.py
```diff
@@ -88,6 +88,116 @@ def _stats(
...
+    categories: Optional[List[float]] = None,
+    percentiles: List[int] = [2, 98],
+    **kwargs: Any,
+) -> List[Dict[Any, Any]]:
+    """Calculate array statistics.
+
+    Args:
+        data (numpy.ndarray): Input array data to get the stats from.
+        categorical (bool): treat input data as categorical data. Defaults to False.
+        categories (list of numbers, optional)
```
vincentsarago, line 124: todo

rio_tiler/utils.py
```diff
@@ -88,6 +88,116 @@ def _stats(
...
+            'percentiles': [38, 147],
+            'min': 20,
+            'max': 180,
+            'std': 28.123562304138662,
+            'histogram': [
+                [1625, 219241, 28344, 15808, 12325, 10687, 8535, 7348, 4656, 1208],
+                [20.0, 36.0, 52.0, 68.0, 84.0, 100.0, 116.0, 132.0, 148.0, 164.0, 180.0]
+            ],
+            'valid_percent': 0.5
+        }
```
vincentsarago, line 142: todo: update

rio_tiler/utils.py
```diff
@@ -88,6 +88,116 @@ def _stats(
...
+
+        if categorical:
+            out_dict = dict(zip(keys.tolist(), counts.tolist()))
+            h_keys = (
+                numpy.array(categories).astype(keys.dtype) if categories else keys
+            ).tolist()
+            histogram = [
+                [out_dict[x] for x in h_keys],
+                h_keys,
+            ]
```
vincentsarago, line 171: if the data is set as `categorical` the histogram is then set to be the count of each values.

rio_tiler/io/base.py
```diff
@@ -917,6 +992,44 @@ def _reader(band: str, *args, **kwargs) -> Metadata:
         }
         return Metadata(**meta)
 
+    def statistics(
+        self,
+        bands: Union[Sequence[str], str] = None,
+        indexes: int = 1,
```
vincentsarago, line 998: make it clear that only the first internal band would be used. but we still allow indexes (e.g if you have band files with multiple internal band you could select the one you want) (this was already possible but we make it clearer by adding it in the function definition

rio_tiler/io/base.py
```diff
@@ -368,7 +397,7 @@ def parse_expression(self, expression: str) -> Tuple:
         return tuple(set(re.findall(_re, expression)))
 
     def info(  # type: ignore
-        self, assets: Union[Sequence[str], str] = None, *args, **kwargs: Any
+        self, assets: Union[Sequence[str], str] = None, **kwargs: Any
```
vincentsarago, line 400: args is not part of the method definition

rio_tiler/io/base.py
```diff
@@ -527,6 +600,7 @@ def _reader(asset: str, *args: Any, **kwargs: Any) -> ImageData:
         if expression:
             blocks = expression.split(",")
             output.data = apply_expression(blocks, assets, output.data)
+            output.band_names = blocks
```
vincentsarago, line 603: here we set the `band_names` to be the name of the expression blocks. This will lose any information about indexes or expression within assets. e.g if you pass `expression="red/blue", indexes=2` or `expression="red/blue", asset_expression=b1*5`, `band_names` will be set to `red/blue`

vincentsarago: `expression` will be mostly used for STAC with `one band` so I think for now it's ok

vincentsarago:
This is ready for review.

Main changes to review are:
- new statistics class and method (matching the one we use in TiTiler). The statistics method only return stats for the whole dataset (using the `preview` method internally to reduce data transfer). In the documentation I'll add notes to explain how to extend the class and create custom statistics methods.

- add `band_names` in ImageData. names are `string` values in form of:
     - COGReader: `["1", "2", "3"...]` (string version of internal rasterio index names)
     - MultiBaseReader: `["{asset}_1", "{asset}_2", "{asset1}_3"]` (asset name + string version of internal rasterio index names)
     - MultiBandReader: `["{band1}", "{band2}", "{band3}"]`. Note we could have add index name but by definition MultiBandReader should be use for per file band data, so there should be only 1 index.

PR #429, opened 2021-09-29
allow non-earth dataset

### LINKED ISSUE #428: CRS Transformation Issue
While testing TiTiler with an S3 hosted COG and I am running into some (not unexpected) issues with the fact that the data are of another body (Europa in this case. Also tested Mars).

I am seeing the following error when attempting to hit the /cog/bounds endpoint.

```
File "rasterio/_err.pyx", line 215, in rasterio._err.exc_wrap_pointer
rasterio._err.CPLE_NotSupportedError: Cannot find coordinate operations from 'PROJCRS["Equirectangular EUROPA",BASEGEOGCRS["GCS_EUROPA",DATUM["D_EUROPA",ELLIPSOID["EUROPA_localRadius",1560800,0,LENGTHUNIT["metre",1,ID["EPSG",9001]]]],PRIMEM["Reference_Meridian",0,ANGLEUNIT["degree",0.0174532925199433,ID["EPSG",9122]]]],CONVERSION["unnamed",METHOD["Equidistant Cylindrical (Spherical)",ID["EPSG",1029]],PARAMETER["Latitude of 1st standard parallel",0,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8823]],PARAMETER["Longitude of natural origin",180,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8802]],PARAMETER["False easting",0,LENGTHUNIT["metre",1],ID["EPSG",8806]],PARAMETER["False northing",0,LENGTHUNIT["metre",1],ID["EPSG",8807]]],CS[Cartesian,2],AXIS["easting",east,ORDER[1],LENGTHUNIT["metre",1,ID["EPSG",9001]]],AXIS["northing",north,ORDER[2],LENGTHUNIT["metre",1,ID["EPSG",9001]]]]' to 'EPSG:4326'
```

Speaking with the [devs over on that project](https://github.com/developmentseed/titiler/discussions/375) it sounds like this could be an issues with [how the bounds are being computed](https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/cogeo.py#L103-L105). The morecantile project pulls a [transformation object](https://github.com/developmentseed/morecantile/blob/master/morecantile/models.py#L141-L155) from pyproj. I wonder if a similar solution might be viable here so that the coordinate transformation can be generated?

Also, here is the direct link to the data if anyone is interested! https://asc-jupiter.s3-us-west-2.amazonaws.com/europa/galileo_voyager/controlled_mosaics/11ESCOLORS01-02_GalileoSSI_Equi-cog.tif

vincentsarago:
closes #428 

This PR adds a major breaking change in the base classes by replacing `bounds` by `dataset_bounds` and `dataset_crs`. 
The idea is to stop storing the bounds in WGS84 but keep the original CRS. We add a `bounds` property to the base class which will then translate the `dataset_bounds` to wgs84 and allow errors by returning default (-180, -90, 180, 90).

Same for `min/max` zoom definition. if the transformation between input CRS and TMS crs is not possible COGReader will default to TMS min/max zoom (note: the min/max zoom are informative, the `Tile` method won't work anyway.)


rio_tiler/io/base.py
```diff
@@ -34,28 +36,31 @@ class SpatialMixin:
...
-    @property
-    def center(self) -> Tuple[float, float, int]:
-        """Dataset center + minzoom."""
-        return (
-            (self.bounds[0] + self.bounds[2]) / 2,
-            (self.bounds[1] + self.bounds[3]) / 2,
-            self.minzoom,
-        )
+    dataset_bounds: BBox = attr.ib(init=False)
+    dataset_crs: CRS = attr.ib(init=False)
```
vincentsarago, line 43: Those 2 properties should be set by the users

rio_tiler/io/base.py
```diff
@@ -34,28 +36,31 @@ class SpatialMixin:
...
+            )
+
+        except:  # noqa
+            # If we can't calculate the transformation between dataset CRS and WGS84
+            # bounds default to -180.0, -90.0, 180.0, 90.0
+            warnings.warn(
+                "Cannot dertermine bounds in WGS84 crs, will default to (-180.0, -90.0, 180.0, 90.0).",
+                UserWarning,
+            )
+            return (-180.0, -90.0, 180.0, 90.0)
```
vincentsarago, line 63: we will always default to (-180.0, -90.0, 180.0, 90.0) even if `dataset_bounds` or `dataset_crs` are not set 🤷‍♂️

rio_tiler/io/base.py
```diff
@@ -69,13 +74,27 @@ def tile_exists(self, tile_x: int, tile_y: int, tile_z: int) -> bool:
...
+            return True
+
+        tile_bounds = self.tms.xy_bounds(Tile(x=tile_x, y=tile_y, z=tile_z))
+
+        dataset_bounds = transform_bounds(
+            self.dataset_crs,
+            self.tms.rasterio_crs,
+            *self.dataset_bounds,
+            densify_pts=21,
+        )
```
vincentsarago, line 91: we now compare `bounds` in TMS crs (we were doing it in WGS84 before)

rio_tiler/io/base.py
```diff
@@ -806,7 +825,11 @@ def _reader(band: str, **kwargs: Any) -> Info:
 
         bands_metadata = multi_values(bands, _reader, *args, **kwargs)
 
-        meta = self.spatial_info.dict()
+        meta = {
+            "bounds": self.bounds,
+            "minzoom": self.minzoom,
+            "maxzoom": self.maxzoom,
+        }
```
vincentsarago, line 832: I've decided to remote the `spatial_info` method because it was kinda useless. I also removed `center` property.

rio_tiler/io/cogeo.py
```diff
@@ -427,15 +438,16 @@ def preview(
...
+            bounds=self.dataset_bounds,
+            crs=self.dataset_crs,
             assets=[self.filepath],
         )
 
     def point(
         self,
         lon: float,
         lat: float,
+        coord_crs: CRS = WGS84_CRS,
```
vincentsarago, line 450: surface `coord_crs` here for documentation purpose

rio_tiler/io/base.py
```diff
@@ -27,35 +29,35 @@ class SpatialMixin:
...
+        minzoom (int): Dataset Min Zoom level. **READ ONLY attribute**.
+        maxzoom (int): Dataset Max Zoom level. **READ ONLY attribute**.
+        bounds (tuple): Dataset bounds (left, bottom, right, top). **READ ONLY attribute**.
+        crs (rasterio.crs.CRS): Dataset crs. **READ ONLY attribute**.
 
     """
 
     tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
+    minzoom: int = attr.ib(default=None)
+    maxzoom: int = attr.ib(default=None)
```
vincentsarago, line 41: we put Min/Max zoom in `init` for visibility (**breaking**)

rio_tiler/io/base.py
```diff
@@ -27,35 +29,35 @@ class SpatialMixin:
...
+
     bounds: BBox = attr.ib(init=False)
-    minzoom: int = attr.ib(init=False)
-    maxzoom: int = attr.ib(init=False)
+    crs: CRS = attr.ib(init=False)
```
vincentsarago, line 44: Object created From this baseclass HAVE to set `bounds` and `crs`

rio_tiler/io/base.py
```diff
@@ -27,35 +29,35 @@ class SpatialMixin:
...
 
     @property
-    def center(self) -> Tuple[float, float, int]:
-        """Dataset center + minzoom."""
-        return (
-            (self.bounds[0] + self.bounds[2]) / 2,
-            (self.bounds[1] + self.bounds[3]) / 2,
-            self.minzoom,
-        )
+    def geographic_bounds(self) -> BBox:
```
vincentsarago, line 47: NAME is `To Be Discussed`

rio_tiler/io/base.py
```diff
@@ -69,13 +71,25 @@ def tile_exists(self, tile_x: int, tile_y: int, tile_z: int) -> bool:
...
+            dataset_bounds = transform_bounds(
+                self.crs, self.tms.rasterio_crs, *self.bounds, densify_pts=21,
+            )
+        except:  # noqa
+            # HACK: gdal will first throw an error for invalid transformation
+            # but if retried it will then pass.
+            # Note: It might return `+/-inf` values
+            dataset_bounds = transform_bounds(
+                self.crs, self.tms.rasterio_crs, *self.bounds, densify_pts=21,
+            )
```
vincentsarago, line 86: it's safer to check coordinates in **TMS** projection. BUT for some reason GDAL/PROJ can error on a first attempt to translate coordinates which overflow the CRS `area_of_use` (should return inf or -inf), on a second attempt it will pass 🤷‍♂️

rio_tiler/io/base.py
vincentsarago: Nope, I removed this
1. not related to the goal of the PR
2. unnecessary breaking change

vincentsarago:
@geospatial-jeff @kylebarron let me know if I can help you for the review

PR #455, opened 2021-11-10
refactor SpatialMixin.tile_exists to avoid coordinates overflow

vincentsarago:
Yet another of the `tile_exists` method. 

When converting bounds from the dataset's CRS to the TMS's CRS it could result in coordinates overflow (a dataset can have bounds bigger than the TMS's CRS bounds). It's safer to apply the opposite: converting the tile_bounds from the TMS's CRS to the dataset's CRS.

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -6,6 +6,7 @@ import warnings
 from typing import Any, Coroutine, Dict, List, Optional, Sequence, Tuple, Type, Union
 
 import attr
+import numpy
 from morecantile import Tile, TileMatrixSet
 from rasterio.crs import CRS
 from rasterio.warp import transform_bounds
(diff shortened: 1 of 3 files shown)


tests/test_io_cogeo.py
```diff
@@ -760,6 +761,12 @@ def test_fullEarth():
         img = cog.tile(127, 42, 7, tilesize=64)
         assert img.data.shape == (1, 64, 64)
 
+    with COGReader(
+        COG_EARTH, tms=morecantile.tms.get("EuropeanETRS89_LAEAQuad")
+    ) as cog:
+        img = cog.tile(0, 0, 1, tilesize=64)
+        assert img.data.shape == (1, 64, 64)
```
vincentsarago, line 768: This would have failed with the previous code

rio_tiler/io/base.py
```diff
@@ -74,31 +75,37 @@ def tile_exists(self, tile_x: int, tile_y: int, tile_z: int) -> bool:
             bool: True if the tile intersects the dataset bounds.
 
         """
+        # bounds in TileMatrixSet's CRS
         tile_bounds = self.tms.xy_bounds(Tile(x=tile_x, y=tile_y, z=tile_z))
 
+        # Transform the bounds in the dataset's CRS
```
vincentsarago, line 81:
```suggestion
        # Transform the bounds to the dataset's CRS
```

PR #460, opened 2021-11-17
add rio_tiler.types and fix type hint issues

vincentsarago:
This PR does:
- better define types of the ColorMap objects
- fix type issues for numpy masked array
- create `rio_tiler.types` submodule to store custom type definition

### DIFF
--- a/rio_tiler/colormap.py
+++ b/rio_tiler/colormap.py
@@ -8,13 +8,13 @@ from typing import Dict, List, Sequence, Tuple, Union
 import attr
 import numpy
 
-from .constants import NumType
 from .errors import (
     ColorMapAlreadyRegistered,
     InvalidColorFormat,
     InvalidColorMapName,
     InvalidFormat,
 )
+from .types import ColorMapType, DataMaskType, GDALColorMapType, IntervalColorMapType
 
 try:
     from importlib.resources import files as resources_files  # type: ignore
@@ -23,7 +23,7 @@ except ImportError:
     from importlib_resources import files as resources_files  # type: ignore
 
 
-EMPTY_COLORMAP: Dict = {i: [0, 0, 0, 0] for i in range(256)}
+EMPTY_COLORMAP: GDALColorMapType = {i: (0, 0, 0, 0) for i in range(256)}
 
 DEFAULT_CMAPS_FILES = {
     f.stem: str(f) for f in (resources_files(__package__) / "cmap_data").glob("*.npy")  # type: ignore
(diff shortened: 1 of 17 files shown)


rio_tiler/constants.py
```diff
@@ -2,16 +2,11 @@
...
 
 import morecantile
 from rasterio.crs import CRS
 
-NumType = Union[float, int]
-BBox = Tuple[float, float, float, float]
-ColorTuple = Tuple[int, int, int, int]
-NoData = Union[float, int, str]
-Indexes = Union[Sequence[int], int]
+from .types import BBox, ColorTuple, Indexes, NoData, NumType  # noqa
```
vincentsarago, line 9: I love breaking change but I like to be nice sometimes!

rio_tiler/io/base.py
```diff
@@ -694,12 +695,12 @@ def _reader(asset: str, *args, **kwargs: Any) -> Dict:
 
         data = multi_values(assets, _reader, lon, lat, **kwargs)
 
-        values = [d for _, d in data.items()]
+        values = numpy.array([d for _, d in data.items()])
```
vincentsarago, line 698: from #411

kylebarron (approved):
❤️  types

rio_tiler/io/base.py
kylebarron: oh I forgot about this PR 😄

rio_tiler/colormap.py
```diff
@@ -53,16 +54,14 @@ def _remove_value(cmap: Dict, idx: Sequence[int]) -> None:
         cmap.pop(i, None)
 
 
-def _update_cmap(cmap: Dict, values: Dict) -> None:
+def _update_cmap(cmap: GDALColorMapType, values: GDALColorMapType) -> None:
     """Update a colormap dict."""
     for i, color in values.items():
-        if len(color) == 3:
-            color += [255]
         cmap[i] = color
```
vincentsarago, line 60: Length is defined by the type 🤷

rio_tiler/colormap.py
```diff
@@ -74,19 +73,17 @@ def make_lut(colormap: Dict) -> numpy.ndarray:
     """
     lut = numpy.zeros(shape=(256, 4), dtype=numpy.uint8)
     for i, color in colormap.items():
-        lut[int(i)] = color
+        lut[int(i)] = numpy.array(color)
```
vincentsarago, line 76: not necessary but I think it's help understanding that we put the tuple in the numpy array

rio_tiler/colormap.py
```diff
@@ -293,7 +290,7 @@ def get(self, name: str) -> Dict:
             colormap = numpy.load(cmap)
             assert colormap.shape == (256, 4)
             assert colormap.dtype == numpy.uint8
-            return {idx: value.tolist() for idx, value in enumerate(colormap)}
+            return {idx: tuple(value) for idx, value in enumerate(colormap)}  # type: ignore
```
vincentsarago, line 293: `value` entry in the numpy array could be of any length (which is why we add an `assert` test) but mypy is not happy ;-)

rio_tiler/types.py
```diff
@@ -0,0 +1,27 @@
...
+
+NumType = Union[float, int]
+
+BBox = Tuple[float, float, float, float]
+NoData = Union[float, int, str]
+Indexes = Union[Sequence[int], int]
+
+DataMaskType = Tuple[numpy.ndarray, numpy.ndarray]
+
+ColorTuple = Tuple[int, int, int, int]
```
vincentsarago, line 15:
```suggestion
ColorTuple = Tuple[int, int, int, int]  # (red, green, blue, alpha)
```

PR #476, opened 2022-01-24
AWS_HTTPS support + fix missing scheme

emmanuelmathot:
**Proposed Changes:**

This PR add support for setting the S3 endpoint url scheme via the `AWS_HTTPS` environment variables in `aws_get_object` function using boto3. AWS_HTTPS is aligned on on gdal [vsis3](https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files) implmenation: https://github.com/OSGeo/gdal/blob/12aaf4be2c20a399efe005482e811a579a12a9fb/port/cpl_aws.cpp#L1207

**PR Checklist:**

- [X] Tests pass (run `tox`)

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -42,7 +42,7 @@ def aws_get_object(
         session = boto3_session()
         # AWS_S3_ENDPOINT and AWS_HTTPS are GDAL config options of vsis3 driver
         # https://gdal.org/user/virtual_file_systems.html#vsis3-aws-s3-files
-        endpoint_url = os.environ.get("AWS_S3_ENDPOINT", None) 
+        endpoint_url = os.environ.get("AWS_S3_ENDPOINT", None)
         if endpoint_url is not None:
             use_https = os.environ.get("AWS_HTTPS", "YES")
             if use_https.upper() in ["YES", "TRUE", "ON"]:
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -41,7 +41,7 @@ def aws_get_object(
     if not client:
         session = boto3_session()
         endpoint_url = os.environ.get("AWS_S3_ENDPOINT", None)
-        if ( endpoint_url != None ):
+        if ( endpoint_url is not None ):
             use_https = os.environ.get("AWS_HTTPS", "YES")
             if ( use_https == "YES" ):
                 endpoint_url = "https://" + endpoint_url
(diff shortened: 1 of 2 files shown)


rio_tiler/utils.py
```diff
@@ -41,6 +41,12 @@ def aws_get_object(
     if not client:
         session = boto3_session()
         endpoint_url = os.environ.get("AWS_S3_ENDPOINT", None)
+        if endpoint_url is not None:
+            use_https = os.environ.get("AWS_HTTPS", "YES")
+            if use_https == "YES":
```
vincentsarago, line 46:
```suggestion
            if use_https.upper() in ["YES", "TRUE", "ON"]:
```

This better align with GDAL bool

rio_tiler/utils.py
```diff
@@ -41,6 +41,12 @@ def aws_get_object(
     if not client:
         session = boto3_session()
         endpoint_url = os.environ.get("AWS_S3_ENDPOINT", None)
+        if endpoint_url is not None:
```
vincentsarago, line 44: we should add some comments at least in the code saying that `AWS_S3_ENDPOINT` and `AWS_HTTPS` are GDAL configuration options.

vincentsarago:
@emmanuelmathot I'm not quite sure we should use `AWS_S3_ENDPOINT` variable here. It might be better to use a rio-tiler specific variable to not mix GDAL/Boto3 config. 

I'm not sure of this but in theory a STAC url could need an S3 endpoint while the asset could not 🤷

emmanuelmathot:
I agree with a more dedicated solution for S3 config. I came up with this pragmatic one to make it work seamlessly with GDAL config.
Concerning multi S3 endpoints, this is indeed a legitimate use case but unfortunately not trivial to tackle. For the current being, we assume that a deployment is "linked" to a S3 endpoint.
This will remain the case until the S3 URLs do not include the endpoint and the bucket consistently.
Some links for the discussion:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html
https://aws.amazon.com/fr/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
https://aws.amazon.com/fr/s3/features/multi-region-access-points/

rio_tiler/utils.py
emmanuelmathot: Is it what you had in mind?

vincentsarago: yes

vincentsarago:
ok lets go ahead with what you propose. Later we could add another variable if a user need to have different endpoints for the item and assets

PR #478, opened 2022-02-04
add `merged_statistics` for multiBaseReader

vincentsarago:
ref: https://github.com/developmentseed/titiler/discussions/429

In `rio-tiler` v3 we removed the `expression` option in the `statistics` for better consistency (to return Per Assets statistics). We couldn't have `between asset` expression here because the `model` would have been different:

- per asset stats: `Dict[str, Dict[str, BandStatistics]]` -> `{"asset1": {"1": rio_tiler.models.BandStatistics, ...}}`
- merged assets stats: `Dict[str, BandStatistics]]` -> `{"asset1_1": rio_tiler.models.BandStatistics, ...}`

Note: Using `expression` within the MultiBaseReader is really tricky because I guess most of the use case is to use it when STAC is hosting `band per file` dataset (only one band per COG) for which we have specifically design the `MultiBandReader`. The user should be aware that merged assets expression might fail if the assets don't have the same number of bands.

Ideally we could require the expression to be in form of `asset1_b1+asset2_b1` but sadly it's not possible because we are using the `expression` to get names of the `assets` we need to fetch (here https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/base.py#L374-L378) so adding a suffix to the asset name will make the code instable IMO

@geospatial-jeff @kylebarron I'm happy to discuss this a bit more

cc @giswqs

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -452,6 +452,70 @@ class MultiBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
 
         return multi_values(assets, _reader, **kwargs)
 
+    def merged_statistics(  # type: ignore
+        self,
+        assets: Union[Sequence[str], str] = None,
+        expression: Optional[str] = None,
+        asset_indexes: Optional[Dict[str, Indexes]] = None,  # Indexes for each asset
+        asset_expression: Optional[Dict[str, str]] = None,  # Expression for each asset
+        categorical: bool = False,
+        categories: Optional[List[float]] = None,
+        percentiles: List[int] = [2, 98],
+        hist_options: Optional[Dict] = None,
+        max_size: int = 1024,
+        **kwargs: Any,
+    ) -> Dict[str, BandStatistics]:
+        """Return array statistics for multiple assets.
+
+        Args:
+            assets (sequence of str or str): assets to fetch info from.
+            expression (str, optional): rio-tiler expression for the asset list (e.g. asset1/asset2+asset3).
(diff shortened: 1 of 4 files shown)


rio_tiler/io/base.py
```diff
@@ -452,6 +452,70 @@ def _reader(asset: str, *args, **kwargs) -> Dict:
...
+                )
+            assets = assets or self.assets
+
+        data = self.preview(
+            assets=assets,
+            expression=expression,
+            asset_indexes=asset_indexes,
+            asset_expression=asset_expression,
+            max_size=max_size,
+            **kwargs,
```
vincentsarago, line 501: to get stats for a merged stac data we use `self.preview` to first get a merged array for multiple assets

tests/test_io_stac.py
```diff
@@ -429,6 +429,59 @@ def test_statistics_valid(rio):
...
+        assert isinstance(stats["red_1"], BandStatistics)
+
+        # Check Expression
+        stats = stac.merged_statistics(expression="green/red")
+        assert isinstance(stats["green/red"], BandStatistics)
+
+        # Check that we can use expression and asset_expression
+        stats = stac.merged_statistics(
+            expression="green/red",
+            asset_expression={"green": "b1*2", "red": "b1+100"},
```
vincentsarago, line 480: we can still apply `per asset` expression or indexes in addition to the expression

giswqs:
Thank you for adding this feature. Much appreciated.

PR #496, opened 2022-04-24
Fixed markdown formatting (line 48)

kant:


vincentsarago (approved):
🙏 thanks

docs/src/mosaic.md
```diff
@@ -45,7 +45,7 @@ Inputs:
 - **chunk_size**: optional, control the number of assets to process per loop.
 - **threads**: optional, number of threads to use in each loop.
 - **allowed_exceptions**: optional, allow some exceptions to be ignored.
-- **\**kwargs**: tiler specific keyword arguments.
+- **\*kwargs**: tiler specific keyword arguments.
```
kylebarron, line 48: I think you want this instead:
```md
- **\*\*kwargs**: tiler specific keyword arguments.
```
Right now kwargs only shows up with one literal asterisk:
image

PR #498, opened 2022-05-25
enable GCPs dataset in COGReader

vincentsarago:
This PR does:
- Deprecate `GCPCOGReader` 
- enable GPCs dataset in COGReader

cc @lossyrob

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -119,9 +119,19 @@ class COGReader(BaseReader):
         if self.post_process is not None:
             self._kwargs["post_process"] = self.post_process
 
-        self.dataset = self.dataset or self._ctx_stack.enter_context(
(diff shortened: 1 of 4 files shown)


rio_tiler/io/cogeo.py
```diff
@@ -119,9 +119,19 @@ def __attrs_post_init__(self):
         if self.post_process is not None:
             self._kwargs["post_process"] = self.post_process
 
-        self.dataset = self.dataset or self._ctx_stack.enter_context(
-            rasterio.open(self.input)
-        )
+        if not self.dataset:
+            dataset = self._ctx_stack.enter_context(rasterio.open(self.input))
```
vincentsarago, line 123: We first open the input file (but register it in the `ctx_stack` so it will be automatically closed when we close the COG instance

PR #515, opened 2022-08-29
reader functions returns ImageData object

vincentsarago:
ref #513


rio_tiler/io/cogeo.py
```diff
@@ -429,19 +429,16 @@ def part(
         if expression and indexes:
             blocks = get_expression_blocks(expression)
             bands = [f"b{bidx}" for bidx in indexes]
-            data = apply_expression(blocks, bands, data)
-
-        if bounds_crs and bounds_crs != dst_crs:
-            bbox = transform_bounds(bounds_crs, dst_crs, *bbox, densify_pts=21)
+            img.data = apply_expression(blocks, bands, img.data)
```
vincentsarago, line 432: we might move `apply_expression` in ImageData class later

PR #519, opened 2022-08-29
generalize buffer option

vincentsarago:
ref #514


rio_tiler/reader.py
```diff
@@ -204,6 +215,32 @@ def part(
                 width = max_size
                 height = math.ceil(width * ratio)
 
+    if buffer:
```
vincentsarago, line 218: This is a totally different approach for applying buffer. 

The main idea is that we first need to know the output `width` and `height` before adding a `buffer`. For `tile` it's easier because we know the output size (256x256) before calling the `part` method. 

The change proposed here assumed there is a linear link between the output size and the vrt size so we first apply the buffer to `height` and `width` then go back and calculate the new `vrt width/height` and then derive the new transform. 

We could also do it as we were doing before:
- get bounds for `part` or `tile`
- find the output resolution.
- calculate the new bounds (with buffer pixel)
- re-call `get_vrt_transform(` with the new bounds and output width/height. 

🤷

rio_tiler/reader.py
```diff
@@ -206,6 +215,34 @@ def part(
...
+        )
+
+        # new output size
+        out_height += int(buffer * 2)
+        out_width += int(buffer * 2)
+
+        # re-calculate the transform given the new bounds, height and width
+        vrt_transform, vrt_width, vrt_height = get_vrt_transform(
+            src_dst, bounds, out_height, out_width, dst_crs=dst_crs
+        )
```
vincentsarago, line 242: This approach will be a bit slower because of the call to `get_vrt_transform` but it will produce better result

tests/test_reader.py
```diff
@@ -387,3 +387,58 @@ def test_point():
...
+        assert imgb.width == nx
+        assert imgb.height == ny
+
+    assert numpy.array_equal(img.data, imgb.data)
+    assert img.bounds == imgb.bounds
+
+    # No resampling is involved. Because we read the full resolution data
+    # all arrays should be equal
+    numpy.array_equal(img_no_buffer.data, imgb.data[:, 2:-2, 2:-2])
+    numpy.array_equal(img_no_buffer.data, img.data[:, 2:-2, 2:-2])
```
vincentsarago, line 444: here we make sure that reading with buffer (manually created and using buffer=) give the same result as non-buffer read

rio_tiler/reader.py
```diff
@@ -206,6 +215,34 @@ def part(
...
+
+    window = windows.Window(col_off=0, row_off=0, width=vrt_width, height=vrt_height)
+
     if padding > 0 and not is_aligned(src_dst, bounds, out_height, out_width, dst_crs):
```
vincentsarago, line 246: just a FYI:
padding is different than buffer because it adds pixel around the VRT while buffer adds pixel around the output image

rio_tiler/reader.py
geospatial-jeff: Yeah I'm not sure there is a way around this because we need to run `get_vrt_transform` first before doing these buffer calculations 🤷

geospatial-jeff (approved):

PR #521, opened 2022-08-30
remove gcp reader

geospatial-jeff:


tests/test_io_cogeo.py
```diff
@@ -450,113 +448,6 @@ def callback(data, mask):
...
-                tile_y = 120
-                data, mask = cog.tile(tile_x, tile_y, tile_z)
-                assert data.shape == (1, 256, 256)
-                assert mask.shape == (256, 256)
-            # https://github.com/rasterio/rasterio/issues/2092
-            # assert cog.dataset.closed
-            assert not cog.src_dataset.closed
-        assert cog.src_dataset.closed
-
-    with COGReader(COG_GCPS, nodata=0) as cog:
```
vincentsarago, line 506: we need to keep those tests!

vincentsarago (changes requested):
we need to keep the GCPS file tests

vincentsarago (approved):

Issue #509, opened 2022-08-09
Do not calculate min/max zoom on Open in COGReader


vincentsarago:
https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/cogeo.py#L162-L185

zoom are not often used (only in `info()` method) but the cost of calculating them is quite high

PR #522, opened 2022-08-31
remove min/max zoom from baseclass attribute

vincentsarago:
ref #509 

This PR removes `minzoom` and `maxzoom` from the `SpatialMixin` base class (thus on other classes). The main effect is that now the user can't set `min/max` zoom as attribute in `COGReader`. 

For STACReader (multibase) and `MultiBandReader` those attributes are still available.

### DIFF
--- a/rio_tiler/io/cogeo.py
+++ b/rio_tiler/io/cogeo.py
@@ -156,7 +156,7 @@ class COGReader(BaseReader):
         """Support using with Context Managers."""
         self.close()
 
-    def _get_vrt_info(self):
+    def _dst_geom_in_tms_crs(self):
         """Return dataset info in TMS projection."""
         if self.dataset.crs != self.tms.rasterio_crs:
             dst_affine, w, h = calculate_default_transform(
@@ -181,7 +181,7 @@ class COGReader(BaseReader):
             tilesize = self.tms.tileMatrix[0].tileWidth
 
             try:
-                dst_affine, w, h = self._get_vrt_info()
+                dst_affine, w, h = self._dst_geom_in_tms_crs()
 
                 # The minzoom is defined by the resolution of the maximum theoretical overview level
                 # We assume `tilesize`` is the smallest overview size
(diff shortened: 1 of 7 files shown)


vincentsarago:
This PR should save couple ms when opening a new file (avoid PROJ operations).

rio_tiler/io/cogeo.py
```diff
@@ -198,10 +195,24 @@ def _set_zooms(self):
...
             minzoom, maxzoom = self.tms.minzoom, self.tms.maxzoom
 
-        self.minzoom = self.minzoom if self.minzoom is not None else minzoom
-        self.maxzoom = self.maxzoom if self.maxzoom is not None else maxzoom
+        self._minzoom = minzoom
+        self._maxzoom = maxzoom
         return
 
+    @property
+    def minzoom(self):
```
geospatial-jeff, line 203: It's sort of weird to me that calling `minzoom` or `maxzoom` sets both.  Its probably not a huge deal but it is a side effect that I think we should try to avoid if possible.  Maybe there is a clever way to split the zoom methods into two, where one handles min and the other handles max.

I wonder how expensive `calculate_default_transform` is.  If it's cheap it might make sense to just call it twice, once for min and once for max.  Then we could break this up.

rio_tiler/io/cogeo.py
```diff
@@ -198,10 +195,24 @@ def _set_zooms(self):
...
+        """Return dataset minzoom."""
+        if self._minzoom is None or self._maxzoom is None:
+            self._set_zooms()
```
geospatial-jeff, line 206: This is a problem in rio-tiler 3.0 too but I'm realizing that the `tilesize` argument accepted by `get_zooms` is not passed through which probably isn't a huge deal because 256 is most common, but wouldn't this break if using a TMS with a larger tile size (ex. 512)?

Maybe `tilesize` argument should be set from `self.tms`?

rio_tiler/io/cogeo.py
vincentsarago: Good point, 
`calculate_default_transform` can be expensive but `min/max zoom` level shouldn't be called too often so I think we can split the `get_zooms`. I'll work on this tomorrow!

rio_tiler/io/cogeo.py
vincentsarago: yeah I'm not sure about this. it's kinda weird, basically a TMS can have a tilesize but the user can also set the tilesize to what ever it wants.

I think we should get the tilesize from the TMS tho.

vincentsarago:
thanks @geospatial-jeff I've split the `zoom` calculation method and also confirmed that it doesn't cost us much because rasterio/gdal seem to do some caching so we pay the price of `calculate_default_transform` only once

PR #523, opened 2022-09-06
move apply_expression into ImageData class and use `b{ix}` for band names

vincentsarago:
this PR does:
- add `apply_expression` into `ImageData` class to avoid code duplication (this is possible because we now have a proper ImageData object out the `reader` methods.) 
- add `b` prefix for band names **breaking**. We should have done this from the beginning!!! (e.g if you pass `expression="b1,b2"` band names would have been `'b1', 'b2'` while `indexes=(1, 2)` would have given `('1', '2')`
    ```python
    # before
    with COGReader("cog.tif") as cog:
        img = cog.read()
        print(cog.band_names)
        >>> ["1", "2", "3"]

        print(cog.info().band_metadata)
        >>> [("1", {}), ("2", {}), ("3", {})]

        print(cog.info().band_descriptions)
        >>> [("1", ""), ("2", ""), ("3", "")]

        print(list(cog.statistics()))
        >>> ["1", "2", "3"]

    # now
    with COGReader("cog.tif") as cog:
        img = cog.read()
        print(img.band_names)
        >>> ["b1", "b2", "b3"]

        print(cog.info().band_metadata)
        >>> [("b1", {}), ("b2", {}), ("b3", {})]

        print(cog.info().band_descriptions)
        >>> [("b1", ""), ("b2", ""), ("b3", "")]

        print(list(cog.statistics()))
        >>> ["b1", "b2", "b3"]

    with STACReader("stac.json") as stac:
        print(stac.tile(701, 102, 8, assets=("green", "red")).band_names)
        >>> ["green_b1", "red_b1"]
    ```


rio_tiler/io/base.py
```diff
@@ -459,31 +459,18 @@ def tile(
...
+            expr = asset_expression.get(asset)  # type: ignore
+            idx = asset_indexes.get(asset) or kwargs.pop("indexes", None)  # type: ignore
             with self.reader(url, tms=self.tms, **self.reader_options) as cog:  # type: ignore
-                data = cog.tile(
-                    *args,
-                    indexes=asset_indexes.get(asset, kwargs.pop("indexes", None)),  # type: ignore
-                    expression=asset_expression.get(asset),  # type: ignore
-                    **kwargs,
-                )
+                data = cog.tile(*args, indexes=idx, expression=expr, **kwargs)
```
vincentsarago, line 465: unrelated to the PR goal!

rio_tiler/io/base.py
```diff
@@ -459,31 +459,18 @@ def tile(
...
-            tile_z,
-            **kwargs,
-        )
-
+        img = multi_arrays(assets, _reader, tile_x, tile_y, tile_z, **kwargs)
         if expression:
-            blocks = get_expression_blocks(expression)
-            output.data = apply_expression(blocks, assets, output.data)
-            output.band_names = blocks
+            img.apply_expression(expression, bands=assets)  # type: ignore
```
vincentsarago, line 471: here we overwrite the `bands` that will be use as expression.

FYI:
- we now have `band_names = ["green_b1", "red_b1"]`
- up to now we are assuming that `expression=green/red` mean there is only one band in each dataset **THIS IS BAD**
- in this line we tell `apply_expression` to overwrite the bands_names to `["green", "red"]` to keep the compatibility.
- The proper expression should be `expression="green_b1/red_b1"` but this will be maybe too big of a **breaking change**

@geospatial-jeff any advice here?

tests/test_io_MultiBand.py
```diff
@@ -59,86 +59,88 @@ def _get_band_url(self, band: str) -> str:
 def test_MultiBandReader():
     """Should work as expected."""
     with BandFileReader(PREFIX) as cog:
-        assert cog.bands == ["b1", "b2"]
+        assert cog.bands == ["band1", "band2"]
```
vincentsarago, line 62: changed the file names to make it more explicit

rio_tiler/io/base.py
vincentsarago: opened #524 to talk about this

geospatial-jeff: > The proper expression should be expression="green_b1/red_b1" but this will be maybe too big of a breaking change

I think we should do something like this.  We are breaking things anyways so might as well get it right now.  Representing the full hierarchy of `asset::band` seems like the best way to support applying expressions across multi-band assets.

geospatial-jeff (commented):
Left some thoughts in #524

Issue #485, opened 2022-02-21
Avoid reprojection (WarpedVRT) when not needed


vincentsarago:
when we created rio-tiler we had 2 main requirement:
- get a mercator tile from any raster
- be able to override nodata value (for Landsat data)

Because of this we used rasterio [WarpedVRT](https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/reader.py#L79) which is really convenient. 

Info from https://github.com/OSGeo/gdal/issues/5343, seems to indicate that we should not used WarpedVRT when no reprojection is not needed (for speed and for resampling artefacts).

mdsumner:
fwiw it's certainly possible to use the warper to give exactly what the RasterIO function does, you need an "aligned extent" - the exact numeric equivalent of what the line/col (window offset,size) inputs to RasterIO use - and a nrow/ncol that is an integer divisor of the source, and resample = "near".  

There are advantages because you can have multiple inputs from various sources too. I'm just saying this here as I'm interested in cross-community shared understanding, I realize it's not exactly relevant right here. But, I'm looking for interested allies across R and Python and GDAL here. :)  Finaly, I have a nascent PR-for-GDAL  that I'm planning, I'd like the warper to have a "snap" option - so that it literally does what the analagous call to RasterIO would do (snap to pixel alignment), not snap to exact target extent.

PR #529, opened 2022-09-26
only use WarpedVRT when doing reprojection or nodata overwride

vincentsarago:
ref https://github.com/cogeotiff/rio-tiler/issues/485

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -1,8 +1,9 @@
 """rio-tiler.reader: low level reader."""
 
+import contextlib
 import math
 import warnings
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+from typing import Callable, Dict, Optional, Tuple, Union
 
 import numpy
 from affine import Affine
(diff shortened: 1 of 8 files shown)


rio_tiler/utils.py
```diff
@@ -386,8 +384,6 @@ def linear_rescale(
 def _requested_tile_aligned_with_internal_tile(
     src_dst: Union[DatasetReader, DatasetWriter, WarpedVRT],
     bounds: BBox,
-    height: Optional[int] = None,
-    width: Optional[int] = None,
```
vincentsarago, line 390: they aren't used 🤦

tests/test_io_cogeo.py
```diff
@@ -709,14 +708,13 @@ def test_tile_read_alpha():
     """Read masked area."""
     # non-boundless tile covering the alpha masked part
     with COGReader(COG_ALPHA) as cog:
-        with pytest.warns(AlphaBandWarning):
```
vincentsarago, line 712: removed the AlphaBandWarning check in reader

rio_tiler/reader.py
```diff
@@ -15,29 +16,67 @@
...
     nodata: Optional[NoData] = None,
-    unscale: bool = False,
-    resampling_method: Resampling = "nearest",
     vrt_options: Optional[Dict] = None,
+    resampling_method: Resampling = "nearest",
+    force_binary_mask: bool = True,
+    unscale: bool = False,
     post_process: Optional[
         Callable[[numpy.ndarray, numpy.ndarray], DataMaskType]
     ] = None,
```
vincentsarago, line 82: reordered a bit the inputs for consistency with other functions

rio_tiler/reader.py
```diff
@@ -46,129 +85,182 @@ def read(
...
     """
     if isinstance(indexes, int):
         indexes = (indexes,)
 
-    if indexes is None:
-        indexes = non_alpha_indexes(src_dst) if nodata is None else src_dst.indexes
-        if indexes != src_dst.indexes:
-            warnings.warn(
-                "Alpha band was removed from the output data array", AlphaBandWarning
-            )
```
vincentsarago, line 72: we can't really do this and don't think it's a big deal!

rio_tiler/reader.py
```diff
@@ -46,129 +85,182 @@ def read(
...
+            if vrt_options:
+                vrt_params.update(**vrt_options)
 
-    with WarpedVRT(src_dst, **vrt_params) as vrt:
-        if not window:
-            window = windows.Window(
-                col_off=0, row_off=0, width=vrt.width, height=vrt.height
-            )
+            # TODO: Check if we fetch the Overviews when not using transform
+            dataset = ctx.enter_context(WarpedVRT(src_dst, **vrt_params))
```
vincentsarago, line 138: when we have either dst_crs / nodata / vrt_option we use Warped VRT

rio_tiler/reader.py
```diff
@@ -46,129 +85,182 @@ def read(
...
+            if isinstance(window, tuple):
+                window = windows.Window.from_slices(
+                    *window, height=dataset.height, width=dataset.width, boundless=True
+                )
+            if min(window.col_off, window.row_off) < 0:
+                boundless = True
+
+            wbounds = dataset.window_bounds(window)
+            if wbounds[2] > dataset.bounds[2] or wbounds[3] > dataset.bounds[3]:
+                boundless = True
```
vincentsarago, line 160: boundless is not a allowed for warpedVRT so we can't set it to True automatically. (we could maybe set it to True and then set it to false for WarpedVRT 🤷

rio_tiler/reader.py
```diff
@@ -202,127 +296,121 @@ def part(
...
+            window=window,
+            nodata=nodata,
+            vrt_options=vrt_params,
+            resampling_method=resampling_method,
+            force_binary_mask=force_binary_mask,
+            unscale=unscale,
+            post_process=post_process,
+        )
+
+    # else no re-projection needed
```
vincentsarago, line 354: 👇 is the new Simple dataset Read. Sadly, because of buffer and padding it feels we are duplicating a lot of code ... but I couldn't find a better way!

rio_tiler/reader.py
```diff
@@ -335,65 +423,74 @@ def point(
...
 
     Returns:
         PointData
 
     """
-    if vrt_options:
-        warnings.warn(
-            "'vrt_options' has no influence over the `point()` method.",
-            UserWarning,
-        )
```
vincentsarago, line 351: we can now ;-)

rio_tiler/reader.py
geospatial-jeff: I think we could deduplicate some code but that would require refactoring `read` which doesn't seem like a good idea.  The `if...else` here is easy enough to follow so I don't mind the code duplication too much.

geospatial-jeff (approved):

Issue #488, opened 2022-03-25
Forward statistics metadata in ImageData 


vincentsarago:
When reading a GeoTIFF dataset, we can have access to GDAL statistics dataset (if added by the user). I think it will be great if we could forward those in the ImageData (output of image methods, e.g. tile(), preview()...).

```
STATISTICS_MEAN: mean
STATISTICS_MINIMUM: minimum
STATISTICS_MAXIMUM: maximum
STATISTICS_STDDEV: standard deviation
STATISTICS_APPROXIMATE: only present if GDAL has computed approximate statistics
STATISTICS_VALID_PERCENT: percentage of valid (not nodata) pixel
```
ref: https://gdal.org/user/raster_data_model.html#metadata

If we are able to ship those in the ImageData object then we could maybe use them here https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/models.py#L333 instead of the min/max of the datatype 🤷‍♂️

PR #531, opened 2022-10-03
forward dataset statistics to ImageClass

vincentsarago:
ref #488


rio_tiler/models.py
```diff
@@ -256,7 +256,7 @@ class ImageData:
         crs (rasterio.crs.CRS, optional): Coordinates Reference System of the bounds.
         metadata (dict, optional): Additional metadata. Defaults to `{}`.
         band_names (list, optional): name of each band. Defaults to `["1", "2", "3"]` for 3 bands image.
-
+        dataset_statistics (list, optional): dataset statistics `{"b1": (min,max), "b2": (min,max)}`
```
vincentsarago, line 259: todo: update

rio_tiler/models.py
```diff
@@ -391,6 +406,15 @@ def apply_color_formula(self, color_formula: Optional[str]):
...
         """Apply expression to the image data."""
         blocks = get_expression_blocks(expression)
+
+        stats = self.dataset_statistics
+        if numpy.array(stats).all() is not None:
+            res = []
+            for prod in itertools.product(*stats):  # type: ignore
+                res.append(apply_expression(blocks, self.band_names, numpy.array(prod)))
+
+            stats = list(zip([min(r) for r in zip(*res)], [max(r) for r in zip(*res)]))
```
vincentsarago, line 416: This is the only real interesting part of the addition (because technically you can access the metadata directly with rasterio). Here we calculate the result of the expression on the min/max values so we get the min/max of the expression output

rio_tiler/reader.py
```diff
@@ -188,6 +188,19 @@ def read(
...
+            tags = dataset.tags(ix)
+            stat_min: Optional[float] = None
+            stat_max: Optional[float] = None
+            if all(
+                stat in tags for stat in ["STATISTICS_MINIMUM", "STATISTICS_MAXIMUM"]
+            ):
+                stat_min = float(tags.get("STATISTICS_MINIMUM"))
+                stat_max = float(tags.get("STATISTICS_MAXIMUM"))
+
+            dataset_statistics.append((stat_min, stat_max))
```
vincentsarago, line 202: we only add `dataset_statistics` if all the bands have it

rio_tiler/models.py
```diff
@@ -473,7 +498,7 @@ def render(
                 kwargs.update({"crs": self.crs})
 
         data = self.data.copy()
-        datatype_range = (dtype_ranges[str(data.dtype)],)
+        datatype_range = self.dataset_statistics or (dtype_ranges[str(data.dtype)],)
```
vincentsarago, line 501: if present we use the dataset min/max for the auto rescaling!

geospatial-jeff (approved):
LGTM

PR #533, opened 2022-10-07
handle image without geo metadata

vincentsarago:


rio_tiler/io/cogeo.py
```diff
@@ -685,3 +693,51 @@ def __attrs_post_init__(self):
...
+        ycoord = self.tile_size * t.y * res
+
+        return morecantile.Coords(xcoord, ycoord)
+
+    def xy_bounds(self, *tile: morecantile.Tile) -> morecantile.BoundingBox:
+        """Return the bounding box of the (x, y, z) tile"""
+        t = _parse_tile_arg(*tile)
+        left, bottom = self._ul(t)
+        right, top = self._ul(morecantile.Tile(t.x + 1, t.y + 1, t.z))
+        return morecantile.BoundingBox(left, top, right, bottom)
```
vincentsarago, line 743: Origin of the non-geo image is bottom left so tile bounds is (minx, maxy, maxx, many)

rio_tiler/io/cogeo.py
```diff
@@ -685,3 +693,51 @@ def __attrs_post_init__(self):
...
+
+    def _ul(self, *tile: morecantile.Tile) -> morecantile.Coords:
+        """Return the upper left coordinate of the (x, y, z) tile."""
+        t = _parse_tile_arg(*tile)
+
+        res = 2.0 ** (self.maxzoom - t.z)
+        xcoord = self.tile_size * t.x * res
+        ycoord = self.tile_size * t.y * res
+
+        return morecantile.Coords(xcoord, ycoord)
```
vincentsarago, line 736: Todo: review ☝️

rio_tiler/io/base.py
```diff
@@ -115,11 +125,13 @@ def tile_exists(self, tile_x: int, tile_y: int, tile_z: int) -> bool:
         if not all(numpy.isfinite(tile_bounds)):
             return True
 
+        tile_bounds = _normalize_bounds(tile_bounds)
+        dst_bounds = _normalize_bounds(self.bounds)
```
vincentsarago, line 129: just make sure we alway compare minx.miny,maxx,maxy

rio_tiler/io/cogeo.py
```diff
@@ -119,6 +120,17 @@ def __attrs_post_init__(self):
...
+            warnings.warn(
+                "Dataset has no CRS defined, Will default to Local TMS.", UserWarning
+            )
+            self.geographic_crs = None
+            self.tms = LocalTileMatrixSet(
+                width=self.dataset.width,
+                height=self.dataset.height,
+            )
+            self._minzoom = self.tms.minzoom
+            self._maxzoom = self.tms.maxzoom
```
vincentsarago, line 133: when we don't have dataset with CRS we assume `Local` TMS.

rio_tiler/io/cogeo.py
```diff
@@ -554,11 +566,18 @@ def feature(
...
+        vrt_options = kwargs.pop("vrt_options", {})
+
         # Get BBOX of the polygon
         bbox = featureBounds(shape)
 
+        # If Dataset Origin is top Left (non-geo) we need to invert the bbox
+        # And set src_method="NO_GEOTRANSFORM" to use WarpedVRT
+        if self.bounds[1] > self.bounds[3] and not self.crs:
+            bbox = [bbox[0], bbox[3], bbox[2], bbox[1]]
+            vrt_options.update({"src_method": "NO_GEOTRANSFORM"})
```
vincentsarago, line 578: Kinda weird but we can use WarpedVRT with non geo dataset, we just need to add `src_method="NO_GEOTRANSFORM"` 

`Feature` is the only method that use WarpedVRT for read because of the cutline

vincentsarago:
To be honest I'm not sure what `frontend` application are for this kind of usage 

I can create a tiler doing 
```python

from fastapi import FastAPI, Query
from fastapi.responses import Response
from rio_tiler.io import COGReader

from starlette.middleware.cors import CORSMiddleware


app = FastAPI()
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_credentials=True,
    allow_methods=["GET"],
    allow_headers=["*"],
)

@app.get("/tiles/{z}/{x}/{y}", response_class=Response)
def tile(
    z: int,
    x: int,
    y: int,
    url: str = Query(description="URL"),
):
    with COGReader(url) as dst:
        img = dst.tile(x, y, z)
    content = img.render(img_format="PNG")
    return Response(content, media_type="image/png")


@app.get("/info")
def info(url: str = Query(description="URL")):
    with COGReader(url) as dst:
        return dst.info().dict(exclude_none=True)
```

and using leaflet 

```html
<!doctype html>
<html>
<head>
    <meta charset="utf-8">
    <title>Leaflet Demo</title>
    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.2/dist/leaflet.css"/>
    <style type="text/css">
        html, body, #image2d {
            width: 100%;
            height: 100%;
            padding: 0;
            margin: 0;
        }
    </style>
</head>
<body>
    <div id="image2d"></div>
    <script src="https://unpkg.com/leaflet@1.9.2/dist/leaflet.js"></script>
    <script type="text/javascript">
var map = L.map('image2d', {crs: L.CRS.Simple}).setView(new L.LatLng(0,0), 0);
fetch('http://127.0.0.1:8080/info?url=image.jpg')
    .then(res => {
        if (res.ok) return res.json()
        throw new Error('Network response was not ok.')
    })
    .then(data => {
        dzLayer = L.tileLayer(
            'http://127.0.0.1:8080/tiles/{z}/{x}/{y}?url=image.jpg, {
                minZoom: data.minzoom,
                maxZoom: data.maxzoom
            }
        )
        dzLayer.addTo(map);
        var southWest = map.unproject([0, data.height], map.getMaxZoom());
        var northEast = map.unproject([data.width, 0], map.getMaxZoom());
        map.setMaxBounds(new L.LatLngBounds(southWest, northEast));
    })
    .catch(err => {
        console.warn(err)
    })
    </script>
</body>
</html>

```

🤷

rio_tiler/io/cogeo.py
mdsumner: excellent, this is impossible in some R packages btw which is unhelpful since the warper is a supercharged RasterIO 👌

rio_tiler/io/cogeo.py
```diff
@@ -554,11 +566,18 @@ def feature(
...
             dst_crs = shape_crs
 
+        vrt_options = kwargs.pop("vrt_options", {})
+
         # Get BBOX of the polygon
         bbox = featureBounds(shape)
 
+        # If Dataset Origin is top Left (non-geo) we need to invert the bbox
+        # And set src_method="NO_GEOTRANSFORM" to use WarpedVRT
+        if self.bounds[1] > self.bounds[3] and not self.crs:
```
geospatial-jeff, line 576: I think we could reuse `_normalize_bounds` here.

rio_tiler/io/cogeo.py
```diff
@@ -685,3 +704,41 @@ def __attrs_post_init__(self):
...
+
+
+@attr.s
+class LocalTileMatrixSet:
+    """Fake TMS for local image."""
+
+    width: int = attr.ib()
+    height: int = attr.ib()
+    tile_size: int = attr.ib(default=256)
+    rasterio_crs: CRS = attr.ib(default=None)
```
geospatial-jeff, line 716:
```suggestion
```

rio_tiler/reader.py
```diff
@@ -459,19 +461,26 @@ def point(
...
-            coord_crs, dataset.crs, [coordinates[0]], [coordinates[1]]
-        )
+        dst_coordinates = coordinates
+        if coord_crs != dataset.crs:
+            x, y = transform_coords(
+                coord_crs, dataset.crs, [coordinates[0]], [coordinates[1]]
+            )
+            dst_coordinates = x[0], y[0]
+
+        minx = min(dataset.bounds[0], dataset.bounds[2])
```
geospatial-jeff, line 471: Can reuse `_normalize_bounds` here too

geospatial-jeff (commented):
I do find it a little weird that `COGReader` can accept an image that isn't actually a COG.

rio_tiler/io/cogeo.py
vincentsarago: @geospatial-jeff we need `rasterio_crs` attribute to keep the compatibility with other TMS. https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/cogeo.py#L361

It should always be `None` to match the image dataset CRS (also None)

rio_tiler/io/cogeo.py
vincentsarago: No! because we want the opposite. 

the issue here is that bbox will be in form of `[minx, miny, maxx, maxy]` (output of `featureBounds(shape)`) but we want to bbox to be in form of `[minx, maxy, maxx, miny]` because the origin of non-geo data is the upper left

vincentsarago:
@geospatial-jeff 

> I do find it a little weird that COGReader can accept an image that isn't actually a COG.

Do you mean that you'll prefer to have a specific `Reader` for non-geo image? 


<details>

```python
@attr.s
class ImageReader(COGReader):
    """Image"""

    tms: morecantile.TileMatrixSet = attr.ib(init=False, default=None)
    geographic_crs: CRS = attr.ib(init=False, default=None)

    def __attrs_post_init__(self):
        """Define _kwargs, open dataset and get info."""

        if not self.dataset:
            self.dataset = self._ctx_stack.enter_context(rasterio.open(self.input))

        height, width = self.dataset.height, self.dataset.width
        self.bounds = (0, height, width, 0)
        self.crs = None
        self.transform = transform_from_bounds(*self.bounds, width=width, height=height)

        self.tms = LocalTileMatrixSet(width=width, height=height)
        self._minzoom = self.tms.minzoom
        self._maxzoom = self.tms.maxzoom

        self.nodata = self.dataset.nodata

        if self.colormap is None:
            self._get_colormap()

        if min(
            self.dataset.width, self.dataset.height
        ) > 512 and not self.dataset.overviews(1):
            warnings.warn(
                "The dataset has no Overviews. rio-tiler performances might be impacted.",
                NoOverviewWarning,
            )

    def tile(
        self,
        tile_x: int,
        tile_y: int,
        tile_z: int,
        tilesize: int = 256,
        indexes: Optional[Indexes] = None,
        expression: Optional[str] = None,
        **kwargs: Any,
    ) -> ImageData:
        """Read a Web Map tile from a COG.

        Args:
            tile_x (int): Tile's horizontal index.
            tile_y (int): Tile's vertical index.
            tile_z (int): Tile's zoom level index.
            tilesize (int, optional): Output image size. Defaults to `256`.
            indexes (int or sequence of int, optional): Band indexes.
            expression (str, optional): rio-tiler expression (e.g. b1/b2+b3).
            tile_buffer (int or float, optional): Buffer on each side of the given tile. It must be a multiple of `0.5`. Output **tilesize** will be expanded to `tilesize + 2 * tile_buffer` (e.g 0.5 = 257x257, 1.0 = 258x258). DEPRECATED
            kwargs (optional): Options to forward to the `COGReader.part` method.

        Returns:
            rio_tiler.models.ImageData: ImageData instance with data, mask and tile spatial info.

        """
        if not self.tile_exists(tile_x, tile_y, tile_z):
            raise TileOutsideBounds(
                f"Tile {tile_z}/{tile_x}/{tile_y} is outside {self.input} bounds"
            )

        tile_bounds = self.tms.xy_bounds(morecantile.Tile(x=tile_x, y=tile_y, z=tile_z))

        return self.part(
            tile_bounds,
            height=tilesize,
            width=tilesize,
            max_size=None,
            indexes=indexes,
            expression=expression,
            **kwargs,
        )


    def part(
        self,
        bbox: BBox,
        indexes: Optional[Union[int, Sequence]] = None,
        expression: Optional[str] = None,
        max_size: Optional[int] = None,
        height: Optional[int] = None,
        width: Optional[int] = None,
        **kwargs: Any,
    ) -> ImageData:
        """Read part of a COG.

        Args:
            bbox (tuple): Output bounds (left, bottom, right, top) in target crs ("dst_crs").
            indexes (sequence of int or int, optional): Band indexes.
            expression (str, optional): rio-tiler expression (e.g. b1/b2+b3).
            max_size (int, optional): Limit the size of the longest dimension of the dataset read, respecting bounds X/Y aspect ratio.
            height (int, optional): Output height of the array.
            width (int, optional): Output width of the array.
            kwargs (optional): Options to forward to the `rio_tiler.reader.part` function.

        Returns:
            rio_tiler.models.ImageData: ImageData instance with data, mask and input spatial info.

        """
        kwargs = {**self._kwargs, **kwargs}

        if indexes and expression:
            warnings.warn(
                "Both expression and indexes passed; expression will overwrite indexes parameter.",
                ExpressionMixingWarning,
            )

        if expression:
            indexes = parse_expression(expression)

        window = window_from_bounds(*bbox, transform=self.transform)
        img = reader.read(
            self.dataset,
            window=window,
            max_size=max_size,
            width=width,
            height=height,
            indexes=indexes,
            **kwargs,
        )
        img.assets = [self.input]

        if expression:
            return img.apply_expression(expression)

        return img

    def point(
        self,
        lon: float,
        lat: float,
        indexes: Optional[Indexes] = None,
        expression: Optional[str] = None,
        **kwargs: Any,
    ) -> PointData:
        """Read a pixel value from a COG.

        Args:
            lon (float): Longitude.
            lat (float): Latitude.
            coord_crs (rasterio.crs.CRS, optional): Coordinate Reference System of the input coords. Defaults to `epsg:4326`.
            indexes (sequence of int or int, optional): Band indexes.
            expression (str, optional): rio-tiler expression (e.g. b1/b2+b3).
            kwargs (optional): Options to forward to the `rio_tiler.reader.point` function.

        Returns:
            PointData

        """
        kwargs = {**self._kwargs, **kwargs}

        if indexes and expression:
            warnings.warn(
                "Both expression and indexes passed; expression will overwrite indexes parameter.",
                ExpressionMixingWarning,
            )

        if expression:
            indexes = parse_expression(expression)

        pt = reader.point(
            self.dataset, (lon, lat), indexes=indexes, coord_crs=None, **kwargs
        )
        pt.assets = [self.input]

        if expression:
            return pt.apply_expression(expression)

        return pt

    def feature(self, shape: Dict) -> ImageData:
        """Read part of a COG defined by a geojson feature."""
        raise NotImplemented

```

</details>


This might be a good idea! This will allow using `ImageReader` even with geo image....

vincentsarago:
let's wait for the other PR. I'll revit this later

vincentsarago:
did another try over https://github.com/cogeotiff/rio-tiler/pull/536

Issue #532, opened 2022-10-05
Rename COGReader -> Reader 


vincentsarago:
`rio_tiler.io.COGReader` is not just for COG but it can be used to read all GDAL Raster dataset! 

As this is a pretty big change I'm ok to add `COGReader = Reader` to keep compatibility

vincentsarago:
We would also rename `rio_tiler.io.cogeo` to `rio_tiler.io.rasterio` (to be inline with stac and xarray

PR #530, opened 2022-09-28
Xarray reader

vincentsarago:
ping @geospatial-jeff @kylebarron @TomAugspurger @rabernat

PR #534, opened 2022-10-13
rename `COGReader` to `Reader` and remove relative import

vincentsarago:
ref #532


rio_tiler/colormap.py
```diff
@@ -8,13 +8,18 @@
 import attr
 import numpy
 
-from .errors import (
+from rio_tiler.errors import (
```
vincentsarago, line 11: I've changed my mind about absolute/relative import. IMO it's much more cleaner with absolute import

rio_tiler/io/__init__.py
```diff
@@ -1,5 +1,8 @@
 """rio-tiler.io"""
 
 from .base import BaseReader, MultiBandReader, MultiBaseReader  # noqa
-from .cogeo import COGReader  # noqa
+from .rasterio import Reader
```
vincentsarago, line 4: In preparation for merging #530, it makes sense to me to rename `cogeo` to `rasterio`

rio_tiler/io/__init__.py
```diff
@@ -1,5 +1,8 @@
...
 from .stac import STACReader  # noqa
+
+# Keep Compatibility with <4.0
+COGReader = Reader
```
vincentsarago, line 8: I believe most user already do `from rio_tiler.io import COGReader` not `from rio_tiler.io.cogeo import COGreader`

PR #535, opened 2022-10-13
remove individual options attributes

vincentsarago:
This PR does:
- remove individual options attributes (nodata, unscale, ...) which are forwarded to the low-level reader and are now replaced with a unique `options={}` 
    ```python
    # before
    with COGReader("cog.tif", nodata=1, resampling_method="bilinear") as cog:
        data = cog.preview()
    # now
    with Reader(COGEO, options={"nodata": 1, "resampling_method": "bilinear"}) as cog:
        data = cog.preview()
    ```

This PR is built on top #534

### DIFF
--- a/rio_tiler/io/rasterio.py
+++ b/rio_tiler/io/rasterio.py
@@ -75,12 +75,16 @@ class Reader(BaseReader):
 
     colormap: Dict = attr.ib(default=None)
 
-    options: Dict[str, Any] = attr.ib(factory=dict)
+    options: reader.Options = attr.ib()
 
     # Context Manager to handle rasterio open/close
     _ctx_stack = attr.ib(init=False, factory=contextlib.ExitStack)
     _minzoom: int = attr.ib(init=False, default=None)
     _maxzoom: int = attr.ib(init=False, default=None)
+
+    @options.default
+    def _options_default(self):
+        return {}
 
     def __attrs_post_init__(self):
         """Define _kwargs, open dataset and get info."""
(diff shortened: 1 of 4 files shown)


geospatial-jeff (approved):
Nice!  I never liked having to remember which arguments went were.  The one downside of this change is code editors won't know what arguments are available.  It should be documented somewhere what the possible arguments are for 'options', if not already.

rio_tiler/io/rasterio.py
```diff
@@ -75,13 +75,17 @@ class Reader(BaseReader):
...
+    options: reader.Options = attr.ib()
 
     # Context Manager to handle rasterio open/close
     _ctx_stack = attr.ib(init=False, factory=contextlib.ExitStack)
     _minzoom: int = attr.ib(init=False, default=None)
     _maxzoom: int = attr.ib(init=False, default=None)
 
+    @options.default
+    def _options_default(self):
+        return {}
```
vincentsarago, line 87: I couldn't get mypy to be happy when putting `attr.ib(factory={})` for defaul

rio_tiler/reader.py
```diff
@@ -23,6 +23,16 @@
 from rio_tiler.utils import get_vrt_transform, has_alpha_band, non_alpha_indexes
 
 
+class Options(TypedDict):
```
vincentsarago, line 26: name?

rio_tiler/reader.py
```diff
@@ -23,6 +23,16 @@
...
+    """Reader Options."""
+
+    nodata: Optional[NoData]
```
geospatial-jeff, line 29: I think we want `NotRequired` here isntead https://stackoverflow.com/questions/72326834/how-to-have-optional-keys-in-typeddict

kylebarron: Also note:

> Note: If you import Required or NotRequired from typing_extensions (on Python <3.11), you should also import TypedDict from typing_extensions.

kylebarron:
> The one downside of this change is code editors won't know what arguments are available

I didn't have time for a close review but I do tend to be wary of this because options dicts or `**kwargs` can feel like black holes at times, where you don't know what values are allowed/disallowed. Maybe with the typeddict it's clear enough

PR #536, opened 2022-10-14
add ImageReader for non-geo images

vincentsarago:
Second try after #533 

This PR does:
- introduce `rio_tiler.io.rasterio.ImageReader` to work either with Non-geo or Geo images in a Non-geo manner (reading part of files in the pixel coordinates system)
- Fix a bug in `reader.point` method (we only want one value for the mask)


geospatial-jeff (approved):
Nice I like this a lot better as it's own class 👍

PR #552, opened 2022-11-10
add rasterio env context for STAC

### LINKED ISSUE #550: make use of file stac-extension `header_size` in MultiBaseReader
in https://github.com/stac-extensions/file there is a `header_size` info which in theory could be used to increase `GDAL_INGESTED_BYTES_AT_OPEN` https://developmentseed.org/titiler/advanced/performance_tuning/#gdal_ingested_bytes_at_open and reduce the number of GET request for accessing the asset.

## How to 

- refactor `MultiBaseReader._get_asset_url` 
https://github.com/cogeotiff/rio-tiler/blob/da7b8b315f53418d6e8b36138f6131020da5e484/rio_tiler/io/base.py#L270-L273

to return a dictionary in form of `{"url": str, "env": Optional[Dict]}`

- The `env` could then be used in https://github.com/cogeotiff/rio-tiler/blob/da7b8b315f53418d6e8b36138f6131020da5e484/rio_tiler/io/base.py#L304-L307

like 
```python
def _reader(asset: str, **kwargs: Any) -> Dict:
    asset_info = self._get_asset_url(asset)
    with rasterio.Env(**asset_info.env):
        with self.reader(asset_info.url, tms=self.tms, **self.reader_options) as cog:  # type: ignore
            return cog.info()
```

cc @geospatial-jeff

vincentsarago:
closes #550

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -1,6 +1,7 @@
 """rio_tiler.io.base: ABC class for rio-tiler readers."""
 
 import abc
+import contextlib
 import re
 import warnings
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
@@ -20,7 +21,7 @@ from rio_tiler.errors import (
 )
 from rio_tiler.models import BandStatistics, ImageData, Info, PointData
 from rio_tiler.tasks import multi_arrays, multi_points, multi_values
-from rio_tiler.types import BBox, Indexes
+from rio_tiler.types import AssetInfo, BBox, Indexes
 from rio_tiler.utils import get_array_statistics, normalize_bounds
 
 
@@ -259,6 +260,8 @@ class MultiBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
 
     assets: Sequence[str] = attr.ib(init=False)
 
+    ctx: Any = attr.ib(init=False, default=contextlib.nullcontext)
+
     def __enter__(self):
         """Support using with Context Managers."""
         return self
@@ -268,7 +271,7 @@ class MultiBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
         pass
 
     @abc.abstractmethod
-    def _get_asset_url(self, asset: str) -> str:
+    def _get_asset_info(self, asset: str) -> AssetInfo:
         """Validate asset name and construct url."""
         ...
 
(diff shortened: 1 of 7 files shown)


rio_tiler/io/base.py
```diff
@@ -302,9 +305,12 @@ def info(
             assets = (assets,)
 
         def _reader(asset: str, **kwargs: Any) -> Dict:
-            url = self._get_asset_url(asset)
-            with self.reader(url, tms=self.tms, **self.reader_options) as src:  # type: ignore
-                return src.info()
+            asset_meta = self._get_asset_info(asset)
+            print(asset_meta)
```
vincentsarago, line 309:
```suggestion
            print(asset_meta)
```

vincentsarago, line 309:
```suggestion
```

rio_tiler/io/base.py
```diff
@@ -259,6 +260,8 @@ class MultiBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
 
     assets: Sequence[str] = attr.ib(init=False)
 
+    ctx: Any = attr.ib(init=False, default=contextlib.nullcontext)
```
vincentsarago, line 263: in practice `ctx` will be `rasterio.Env` but this is not mandatory and could be let to `contextlib.nullcontext` is the MultiBaseReader class do not use a rasterio based reader

rio_tiler/io/stac.py
```diff
@@ -224,4 +229,11 @@ def _get_asset_url(self, asset: str) -> str:
         if asset not in self.assets:
             raise InvalidAssetName(f"{asset} is not valid")
 
-        return self.item.assets[asset].get_absolute_href()
+        asset_info = self.item.assets[asset]
+        info = AssetInfo(url=asset_info.get_absolute_href())
+
+        if "file:header_size" in asset_info.extra_fields:
+            h = asset_info.extra_fields["file:header_size"]
+            info["env"] = {"GDAL_INGESTED_BYTES_AT_OPEN": 16384 * math.ceil(h / 16384)}
```
vincentsarago, line 237: ✨

vincentsarago:
@geospatial-jeff any feedback on this?

rio_tiler/io/stac.py
geospatial-jeff: Isn't this the same as `{"GDAL_INGESTED_BYTES_AT_OPEN": h}`

geospatial-jeff (approved):

rio_tiler/io/stac.py
vincentsarago: I guess it should

PR #559, opened 2022-12-10
Fixed duplicate use of y coordinate and fixed order of lat and lon in point() in xarray.py

benjaminleighton:
Corrects the slicing of data in generating PointData to use x and y rather than just y. I also believe that the bounds check and rowcol had ds_lat and ds_lon in the wrong place. I'm not sure of the later and mostly determined it emperically testing with "https://pangeo.blob.core.windows.net/pangeo-public/daymet-rio-tiler/na-wgs84.zarr/"

### DIFF
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -410,4 +410,3 @@ class XarrayReader(BaseReader):
             dataset_statistics=stats,
             band_names=band_names,
         )
-
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -335,17 +335,17 @@ class XarrayReader(BaseReader):
         ds_lon, ds_lat = transform_coords(coord_crs, self.crs, [lon], [lat])
 
         if not (
-            (self.bounds[0] < ds_lat[0] < self.bounds[2])
-            and (self.bounds[1] < ds_lon[0] < self.bounds[3])
+            (self.bounds[0] < ds_lon[0] < self.bounds[2])
+            and (self.bounds[1] < ds_lat[0] < self.bounds[3])
         ):
             raise PointOutsideBounds("Point is outside dataset bounds")
 
-        x, y = rowcol(self.input.rio.transform(), ds_lat, ds_lon)
+        y, x = rowcol(self.input.rio.transform(), ds_lon, ds_lat)
 
         band_names = [str(band) for d in self._dims for band in self.input[d].values]
 
         return PointData(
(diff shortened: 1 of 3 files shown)


rio_tiler/io/xarray.py
```diff
@@ -335,17 +335,17 @@ def point(
         ds_lon, ds_lat = transform_coords(coord_crs, self.crs, [lon], [lat])
 
         if not (
-            (self.bounds[0] < ds_lon[0] < self.bounds[2])
-            and (self.bounds[1] < ds_lat[0] < self.bounds[3])
+            (self.bounds[0] < ds_lat[0] < self.bounds[2])
```
vincentsarago, line 338: @benjaminleighton bounds are in form of `xmin, ymin, xmax, ymax` so `self.bounds[0] < ds_lon[0] < self.bounds[2]` is right, I'm not sure why you want to compare latitude (Y) with bounds[0], bounds[1] which are X coordinates?

rio_tiler/io/xarray.py
```diff
@@ -335,17 +335,17 @@ def point(
...
+            and (self.bounds[1] < ds_lon[0] < self.bounds[3])
         ):
             raise PointOutsideBounds("Point is outside dataset bounds")
 
-        x, y = rowcol(self.input.rio.transform(), ds_lon, ds_lat)
+        x, y = rowcol(self.input.rio.transform(), ds_lat, ds_lon)
```
vincentsarago, line 343: rowcol() takes `lon` (x values in coordinate reference system), `lat` (y values in coordinate reference system.) as input and return a tuple of rows and cols 

There is definitely an issue here but the fix should be 
```
y, x = rowcol(self.input.rio.transform(), ds_lon, ds_lat)
```

rio_tiler/io/xarray.py
```diff
@@ -335,17 +335,17 @@ def point(
...
 
         band_names = [str(band) for d in self._dims for band in self.input[d].values]
 
         return PointData(
-            self.input.data[:, y[0], y[0]],
+            self.input.data[:, x[0], y[0]],
```
vincentsarago, line 348: I agree there is a bug here 🤦 

but the fix should be `self.input.data[:, y[0], x[0]],`

vincentsarago:
Thanks for the latest commit @benjaminleighton 

Would be nice if we could make a nice test for this

benjaminleighton:
I have slightly modified and extended the existing test. I hope this look ok? Happy to modify it further as needed.

PR #567, opened 2023-01-18
add apply_colormap

vincentsarago:


rio_tiler/utils.py
```diff
@@ -417,16 +417,18 @@ def render(
 
     if colormap:
         data, alpha = apply_cmap(data, colormap)
+        # We take both the input mask and the alpha from the colormap
+        # if input mask is not provided then we assume output is wanted without alpha band
+        # this can be seen as a bug but at the time of writing we assume it's a feature.
```
vincentsarago, line 422: it's not a bug, it's a feature

Hopefully, most of the people will use the `render` method from the ImageData class which explicitly uses `add_mask` bool

PR #569, opened 2023-01-25
Add support for aws s3 endpoint with access keys

nadavFux:
Chosen env key names are in accordance with gdal's
A guide on repo convention would have helped me create a destination branch of a better practice


vincentsarago:
Thanks for the PR 🙏 

> A guide on repo convention would have helped me create a destination branch of a better practice

Feel free to start a PR to add one 🙏

vincentsarago:
replaced by #613

PR #574, opened 2023-02-06
move to ruff and other linting update

### LINKED ISSUE #568: move to ruff
Every cool projects are moving to Ruff (https://github.com/charliermarsh/ruff) we should be too!

vincentsarago:
closes #568

### DIFF
--- a/docs/scripts/colormap_thumb.py
+++ b/docs/scripts/colormap_thumb.py
@@ -142,7 +142,7 @@ def plot_color_gradients(cmap_category, cmap_list):
 
     axes[0].set_title(cmap_category + " colormaps", fontsize=14)
 
-    for ax, name in zip(axes, cmap_list):
+    for ax, name in zip(axes, cmap_list, strict=True):
         ax.imshow(gradient, aspect="auto", cmap=make_colormap(name))
         ax.text(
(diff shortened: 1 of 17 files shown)


docs/scripts/colormap_thumb.py
```diff
@@ -142,7 +142,7 @@ def plot_color_gradients(cmap_category, cmap_list):
 
     axes[0].set_title(cmap_category + " colormaps", fontsize=14)
 
-    for ax, name in zip(axes, cmap_list):
+    for ax, name in zip(axes, cmap_list, strict=True):
```
vincentsarago, line 145:
```suggestion
    for ax, name in zip(axes, cmap_list):
```

rio_tiler/io/rasterio.py
```diff
@@ -271,7 +271,7 @@ def statistics(
         self,
         categorical: bool = False,
         categories: Optional[List[float]] = None,
-        percentiles: List[int] = [2, 98],
+        percentiles: Optional[List[int]] = None,
```
vincentsarago, line 274: We defaults to the values set in the `get_array_statistics` function. Setting a default here is unnecessary

PR #578, opened 2023-02-28
Automatically expand 2D numpy array to 3D when creating ImageData

### LINKED ISSUE #577: ImageData validator is ran after default
When trying to create a ImageData object from a 2D array without mask it results in 👇 

```python
import numpy
from rio_tiler.models import ImageData

data = numpy.zeros((256,256))

ImageData(data)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 ImageData(data)

File <attrs generated init rio_tiler.models.ImageData>:6, in __init__(self, data, mask, assets, bounds, crs, metadata, band_names, dataset_statistics)
      4     self.mask = mask
      5 else:
----> 6     self.mask = __attr_factory_mask(self)
      7 self.assets = assets
      8 self.bounds = __attr_converter_bounds(bounds)

File rio_tiler/models.py:300, in ImageData._default_mask(self)
    298 @mask.default
    299 def _default_mask(self):
--> 300     return numpy.zeros((self.height, self.width), dtype="uint8") + 255

File rio_tiler/models.py:445, in ImageData.width(self)
    442 @property
    443 def width(self) -> int:
    444     """Width of the data array."""
--> 445     return self.data.shape[2]

IndexError: tuple index out of range
```

while I was expecting 👇 
https://github.com/cogeotiff/rio-tiler/blob/b22e4d20d03d1790452f9afca0651e08d769697e/rio_tiler/models.py#L286-L292

see https://www.attrs.org/en/stable/init.html#order-of-execution

vincentsarago:
closes #577

### DIFF
--- a/rio_tiler/models.py
+++ b/rio_tiler/models.py
@@ -258,6 +258,14 @@ class PointData:
         )
 
 
+def to_3d(data: numpy.ndarray) -> numpy.ndarray:
+    """Makes sure we have a 3D array."""
+    if len(data.shape) < 3:
+        data = numpy.expand_dims(data, axis=0)
+
+    return data
+
+
 @attr.s
 class ImageData:
     """Image Data class.
@@ -274,7 +282,7 @@ class ImageData:
 
     """
 
-    data: numpy.ndarray = attr.ib()
+    data: numpy.ndarray = attr.ib(converter=to_3d)
     mask: numpy.ndarray = attr.ib()
     assets: Optional[List] = attr.ib(default=None)
     bounds: Optional[BoundingBox] = attr.ib(default=None, converter=to_coordsbbox)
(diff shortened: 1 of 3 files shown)


rio_tiler/models.py
```diff
@@ -283,14 +291,6 @@ class ImageData:
     band_names: List[str] = attr.ib()
     dataset_statistics: Optional[Sequence[Tuple[float, float]]] = attr.ib(default=None)
 
-    @data.validator
-    def _validate_data(self, attribute, value):
-        """ImageData data has to be a 3d array in form of (count, height, width)"""
-        if not len(value.shape) == 3:
-            raise ValueError(
-                "ImageData data has to be an array in form of (count, height, width)"
-            )
```
vincentsarago, line 292: It's easier to expand the data than to raise an Error

PR #583, opened 2023-03-15
Update dynamic_tiler.md

samsammurphy:
removing imports that don't seem to be used in this example

btw.. I am not sure where `ImageResponse` is imported from, see

`response_class=ImageResponse`

also.. I am not familiar with FastAPI, trying to learn from example..


vincentsarago:
🙏 
`ImageResponse` should just be`Response`

Issue #585, opened 2023-03-24
Bug in raster cutline when geometry goes over edge


yellowcap:
The cutline behavior gives unexpected results when a geometry goes over the edge.

For a polygon like in the image below, the cutline algorithm will cut too much of the data. The reason seems to be that when computing the cutline coordinates in pixel space, the values are ["snapped" linearly to the edge](https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/utils.py#L538) when they go over the image edge.

This snapping will essentially project the coordinates horizontally (or vertically) onto the edge. In the case of the image below, this implies that the triangle that is cut out is too big (the angle changes).

![tiranlge_example_qgis](https://user-images.githubusercontent.com/901647/227511472-e7cd42dd-7695-4455-a6e8-c1c43e1a4a90.png)

This problem can be fixed by simply skipping the snapping step. I tried this out and GDAL does not complain when the pixel coordinates go over the max size of the image. So I will propose a fix, but maybe there is a reason why the coordinates were limited in the first place that I am not aware of.

Here is hare plots of the cutline mask with and without the snapping mechanism.

Current behavior:

![Figure_1](https://user-images.githubusercontent.com/901647/227512431-d7835483-acf0-47ea-86cf-6dd8042bf29b.png)

Expected behavior (produced by simply skipping ines 538 and 539):

![Figure_2](https://user-images.githubusercontent.com/901647/227512445-57fc0dae-80f2-4015-a309-fba1591fa18b.png)

vincentsarago:
closed in #586

PR #586, opened 2023-03-24
Fix cutline behavior over edge (#585)

yellowcap:
The test does the cutting using the triangle used as illustration in the related issue https://github.com/cogeotiff/rio-tiler/issues/585.

Note that the sum of the mask is hard coded from summing the values while running the new cutline test manually. I tried to come up with a mathematical way to calculate how big the mask should be, but I realized that would take a bit more time than I thought. 

So if it would be important to have a mathematical way to compute to the magic number of `5984` pixels for this mask let me know I can try to compute this number from the geometry. Please advise on this.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -375,7 +375,6 @@ def test_cutline():
     triangle_bounds = featureBounds(triangle_over_image_edge)
 
     with Reader(COG_RGB) as src:
-        from matplotlib import pyplot as plt
 
         cutline = utils.create_cutline(
             src.dataset, triangle_over_image_edge, geometry_crs="epsg:4326"
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -34,6 +34,7 @@ PIX4D_PATH = os.path.join(S3_LOCAL, KEY_PIX4D)
 COG_DST = os.path.join(os.path.dirname(__file__), "fixtures", "cog_name.tif")
 COG_WEB_TILED = os.path.join(os.path.dirname(__file__), "fixtures", "web.tif")
 COG_NOWEB = os.path.join(os.path.dirname(__file__), "fixtures", "noweb.tif")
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -359,6 +360,29 @@ def test_cutline():
...
+                [-104.775389629827075, 38.953472856486307],
+                [-104.775390888988852, 38.953714348778355],
+            ]
+        ],
+    }
+
+    triangle_bounds = featureBounds(triangle_over_image_edge)
+
+    with Reader(COG_RGB) as src:
+        from matplotlib import pyplot as plt
```
vincentsarago, line 378:
```suggestion
```

tests/test_utils.py
```diff
@@ -359,6 +360,30 @@ def test_cutline():
...
+    # https://github.com/cogeotiff/rio-tiler/issues/585
+    triangle_bounds = featureBounds(triangle_over_image_edge)
+    with Reader(COG_RGB) as src:
+        cutline = utils.create_cutline(
+            src.dataset, triangle_over_image_edge, geometry_crs="epsg:4326"
+        )
+        data, mask = src.part(triangle_bounds, vrt_options={"cutline": cutline})
+        assert sum(mask[:, 0]) == 0  # first line
+        assert sum(mask[0, :]) == 0  # first column
+        assert sum(mask[:, -1]) == 0  # last column
```
vincentsarago, line 385: relying on pixel count is a bit tricky because it can changed depending on the platform and GDAL version. Those are more general tests to make sure it work with `boundless` cutline (first and last line should be only `0`

vincentsarago (approved):

tests/test_utils.py
yellowcap: Great, I did have doubts about the pixel count and like this solution a lot! Thanks 😄

PR #587, opened 2023-03-24
Fix mosaic_point_reader docstring

yellowcap:
Update wording that was probably an artifact from copy-pasting the docstring from the function above this one.


vincentsarago:
🙏

Issue #588, opened 2023-03-24
Finish mosaicking when all pixels within a cutline are populated


yellowcap:
The current implementation of the `MosaicMethodBase` will `exit_when_filled` only when all pixels of the array are populated.

When using a cutline, this is never the case. So for methods such as the [`FirstMethod`](https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/mosaic/methods/defaults.py#L10), the algorithm will run through all the scenes or assets that are passed on.

This is not efficient and could be improved.

The following example will run through all the S2 scenes that are returned from the STAC search (limited to 3 here for demo purposes)

```python
import datetime

import httpx
import rasterio
from matplotlib import pyplot as plt
from rasterio.crs import CRS
from rasterio.features import bounds as featureBounds
from rasterio.warp import transform_geom

from rio_tiler.io import STACReader
from rio_tiler.mosaic import mosaic_reader
# from ipyleaflet import Map, basemaps, GeoJSON
from rio_tiler.utils import create_cutline

stac_endpoint = "https://earth-search.aws.element84.com/v0/search"

# Date filter
date_min = "2022-08-01"
date_max = "2022-08-31"

start = datetime.datetime.strptime(date_min, "%Y-%m-%d").strftime("%Y-%m-%dT00:00:00Z")
end = datetime.datetime.strptime(date_max, "%Y-%m-%d").strftime("%Y-%m-%dT23:59:59Z")

# AOI over Bern
aoi = {
    "type": "Feature",
    "properties": {},
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [-151.98538128444676, 137.785308042979636],
                [427.064224021652308, 334.495939510601829],
                [727.671315771469153, 78.218004147854629],
                [1410.617381359762931, 64.365142776895368],
                [1750.012484948266319, -33.990172956915444],
                [1675.207033545085778, -581.178197109808139],
                [1435.552531827489929, -773.732970166142536],
                [873.126360166541872, -869.317713625761712],
                [722.130171223085426, -885.941147270912893],
                [590.527988198972253, -1478.843613947971562],
                [222.041875731454525, -1501.00819214150647],
                [-250.34069701825797, -1215.63924789974476],
                [-530.168496711635839, -297.19453900514236],
                [-151.98538128444676, 137.785308042979636],
            ]
        ],
    },
}
epsg = 21780

query = {
    "collections": [
        "sentinel-s2-l2a-cogs"
[...]

PR #598, opened 2023-04-24
improve cutline handling

yellowcap:
This is another attempt to resolve https://github.com/cogeotiff/rio-tiler/issues/588

The proposed changes ensure that the `FirstMethod` finishes when all pixels over the target geometry is filled.

The main implementation challenge is: the `pixel_selection` class needs to know the target mask to use in `is_done` function.

This means that the input geometry needs to be converted to a mask as numpy array to which the `pixel_selection` needs to have access.

For this case, the cutline `vrt_option` approach does not work, because for gdal cutlines the mask generation happens on the gdal side and the mask is not returned. So the proposed changes do not use the `vrt_option` argument, but rasterized the geometry directly and applies this mask on the python side after requesting pixels.

```python
import datetime

import httpx
import matplotlib.pyplot as plt
from rasterio.crs import CRS
from rasterio.warp import transform_geom

from rio_tiler.io import STACReader
from rio_tiler.mosaic import mosaic_reader

stac_endpoint = "https://earth-search.aws.element84.com/v0/search"

# Date filter
date_min = "2022-08-01"
date_max = "2022-08-31"

start = datetime.datetime.strptime(date_min, "%Y-%m-%d").strftime("%Y-%m-%dT00:00:00Z")
end = datetime.datetime.strptime(date_max, "%Y-%m-%d").strftime("%Y-%m-%dT23:59:59Z")

# Bern large
aoi = {
    "type": "Feature",
    "properties": {},
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [296.138689305058506, -3217.936052156386722],
                [2770.964285295431182, -5479.41461400965818],
                [3496.34420136157496, -8466.273091929069778],
                [2045.584369229287404, -11794.486824467847327],
                [1106.857419026044226, -15506.725218453404523],
                [4435.071151564818138, -16701.46860962116989],
                [10110.10225961170363, -16232.105134519548301],
                [10408.78810740364861, -15336.047591143724276],
                [10238.110480093964725, -12903.891401980770752],
                [9811.416411819765926, -11367.792756193644891],
                [7763.284884103592049, -7271.529700761304412],
                [5629.814542732583504, -4113.993595532210747],
                [3795.030049153516302, -1980.523254161202203],
                [1490.882080472823873, -401.755201546653552],
                [296.138689305058506, -3217.936052156386722],
            ]
        ],
    },
}
epsg = 21780

query = {
    "collections": [
        "sentinel-s2-l2a-cogs"
    ],  # Make sure to query only sentinel-2 COGs collection
    "datetime": f"{start}/{end}",
    "query": {"eo:cloud_cover": {"lt": 90}},
    "intersects": transform_geom(f"epsg:{epsg}", "epsg:4326", aoi["geometry"]),
    "limit": 2,
    "fields": {
        "include": [
            "id",
            "properties.datetime",
            "properties.eo:cloud_cover",
        ],
        "exclude": ["links"],
    },
}

headers = {
    "Content-Type": "application/json",
    "Accept-Encoding": "gzip",
    "Accept": "application/geo+json",
}


data = httpx.post(stac_endpoint, headers=headers, json=query).json()
print(data["context"])

sceneid = [f["id"] for f in data["features"]]
cloudcover = [f["properties"]["eo:cloud_cover"] for f in data["features"]]
dates = [f["properties"]["datetime"][0:10] for f in data["features"]]

items = [feat | {"stac_version": data["stac_version"]} for feat in data["features"]]


def tiler(asset, *args, **kwargs):
    print(asset["id"])
    with STACReader(None, asset) as src:
        img = src.feature(
            shape=kwargs["shape"],
            shape_crs=kwargs["shape_crs"],
            assets=kwargs["assets"],
        )
        print("min/max/shape", img.data.min(), img.data.max(), img.data.shape)

        return img

img, assets_used = mosaic_reader(
    items,
    tiler,
    shape=aoi,
    shape_crs=CRS.from_epsg(21780),
    assets=["B04", "B03", "B02"],
    threads=1,
)

fig = plt.figure(figsize=(30, 10))

ax = fig.add_subplot(1, 2, 1)
img.rescale(in_range=((0, 5000),))
ax.imshow(img.data_as_image())

ax = fig.add_subplot(1, 2, 2)
ax.imshow(img.mask)

plt.show()
```


rio_tiler/io/rasterio.py
```diff
@@ -409,6 +411,19 @@ def part(
...
+                out_shape=(img.height, img.width),
+                transform=img.transform,
+                all_touched=kwargs.get("all_touched", False),
+                default_value=0,
+                fill=1,
+                dtype="uint8",
+            ).astype("bool")
+            img.cutline_mask = cutline_mask
+            img.array.mask = numpy.where(~cutline_mask, img.array.mask, True)
+
```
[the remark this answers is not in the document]
yellowcap: good idea, just updated it accordingly.

rio_tiler/io/rasterio.py
```diff
@@ -545,6 +546,21 @@ def feature(
             **kwargs,
         )
 
+        if shape is not None:
```
vincentsarago, line 549: shape should always be set

rio_tiler/utils.py
```diff
@@ -636,3 +631,15 @@ def normalize_bounds(bounds: BBox) -> BBox:
...
+def validate_shape_input(shape: Dict) -> Dict:
+    """Ensure input shape is valid and reduce features to geometry"""
+
+    if "geometry" in shape:
+        shape = shape["geometry"]
+
+    if not is_valid_geom(shape):
+        raise RioTilerError("Invalid geometry")
+
+    return shape
```
vincentsarago, line 645: 👍

rio_tiler/io/rasterio.py
```diff
@@ -545,6 +546,21 @@ def feature(
...
+            cutline_mask = rasterize(
+                [shape],
+                out_shape=(img.height, img.width),
+                transform=img.transform,
+                all_touched=kwargs.get("all_touched", False),
```
vincentsarago, line 554: 🤔 not sure about this. can you explain?

rio_tiler/io/rasterio.py
yellowcap: of course! 😶‍🌫️

rio_tiler/io/rasterio.py
yellowcap: This is a gdal option for how to rasterize the geometry. From [gdal docs](https://gdal.org/programs/gdal_rasterize.html)

> Enables the ALL_TOUCHED rasterization option so that all pixels touched by lines or polygons will be updated, not just those on the line render path, or whose center point is within the polygon. Defaults to disabled for normal rendering rules.

See also
https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html#rasterio.features.rasterize

I was insecure about the "api" here, we could also add this as an explicit kwarg to the `feature` function.

vincentsarago: yeah I think it will make more sense to have a specific option in the `feature` method.

yellowcap:
For the docs, it seems like aside from the `all_touched` part, there is no api change on the functions. So maybe just adding another example would be a good way to document? 

Either an extension of https://cogeotiff.github.io/rio-tiler/examples/Using-rio-tiler-mosaic/

or a fresh one similar to the snipped in the ticket above.

rio_tiler/utils.py
yellowcap: Made this method private in latest push.  Let me know if you think it should not be private.

docs/src/advanced/feature.md
```diff
@@ -11,27 +11,38 @@ with Reader("my-tif.tif") as cog:
     img: ImageData = cog.feature(geojson_feature, max_size=1024)  # we limit the max_size to 1024
 ```
 
-Under the hood, the `.feature` method uses `GDALWarpVRT`'s `cutline` option and
-the `.part()` method. The below process is roughly what `.feature` does for you.
+Under the hood, the `.feature` method uses rasterio's (`rasterize`)[https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html#rasterio.features.rasterize]
```
yellowcap, line 14: Fix link brackets.

docs/src/advanced/feature.md
```diff
@@ -11,27 +11,38 @@ with Reader("my-tif.tif") as cog:
...
+    # Apply geometry mask to imagery
+    img.array.mask = numpy.where(~cutline, img.array.mask, True)
 ```
 
-Another interesting fact about the `cutline` option is that it can be used with other methods:
+Another interesting way to cut features is to use the GDALWarpVRT's `cutline`
+option with the .part(), .preview(), or .tile() methods:
 
 ```python
 bbox = featureBounds(feat)
```
yellowcap, line 48: Add  `from rio_tiler.utils import create_cutline`

rio_tiler/io/rasterio.py
```diff
@@ -515,23 +517,24 @@ def feature(
             height (int, optional): Output height of the array.
             width (int, optional): Output width of the array.
             buffer (int or float, optional): Buffer on each side of the given aoi. It must be a multiple of `0.5`. Output **image size** will be expanded to `output imagesize + 2 * buffer` (e.g 0.5 = 257x257, 1.0 = 258x258).
+            all_touched (bool, optional): Shape rasterization parameter. If True, all pixels touched the input shape will be burned in. If false, only pixels whose center is within the input shape or that are selected by Bresenham's line algorithm will be burned in.
```
vincentsarago, line 520: The more I read this the more I think we should always set this to true 🤷

yellowcap: That's fine with me. This parameter has an impact on features that are small or thin when compared with a single pixel. I.e. linear features or small areas. So for masking its usually not a big difference either way.

vincentsarago:
👍 all good for me. 

Would you be able to write a note in the changelog 🙏

yellowcap:
Added changelog note

yellowcap:
One sidenote: the [create_cutline](https://github.com/cogeotiff/rio-tiler/blob/dev/rio_tiler/utils.py#L521) utils function is no longer used in this codebase. What to do in this case? Just leave it be, or deprecate?

rio_tiler/io/rasterio.py
yellowcap: I kept this in the latest version, but let me know if you prefer to remove it and default to `True`. Either way is fine for me.

vincentsarago:
@yellowcap why is the `create_cutline` no more used? I think it should still be!

yellowcap:
> @yellowcap why is the `create_cutline` no more used? I think it should still be!

I first left it in, but the problem is that the cutline mask is done in the projection of the source image, so it won't exactly align with the cutline mask done in the target projection. The "finish early" will fail in some of these cases because there are pixels that are masked in the original projection, but not anymore in the target one. So these pixels will always be masked in the target projection.

I thought about this in terms of what it means for data transfer efficiency. My guess is that this will not have implications for data transfers, but not sure. Maybe some COG internal tile requests can be skipped with the cutline vrt_option in place. Do you know if GDAL optimizes that under the hood, i.e. identifying what tiles not to get based on the cutline?

yellowcap:
Also, I just realized that the masking will not fully work for rasters with multiple resolutions (10m vs 20m) in this case. So I'll have to digg into that, the stop early fails when multiple resolutions are included.

yellowcap:
I added another commit that solves the multiple resolution problem. The approach is to only apply the same cutline mask to all assets using the highest resolution rasterization.
https://github.com/cogeotiff/rio-tiler/pull/598/commits/0894fad31c06e067dda298ec778d8c91de2d174f

vincentsarago:
> I thought about this in terms of what it means for data transfer efficiency. My guess is that this will not have implications for data transfers, but not sure. Maybe some COG internal tile requests can be skipped with the cutline vrt_option in place. Do you know if GDAL optimizes that under the hood, i.e. identifying what tiles not to get based on the cutline?

It might (really doubt) but I don't think this will be a major issueQ 👍 

Let's keep `create_cutline` in utils because it might be useful outside the `feature` method (e.g to mask tiles)https://github.com/developmentseed/titiler/discussions/376

rio_tiler/io/rasterio.py
```diff
@@ -501,6 +501,8 @@ def feature(
         height: Optional[int] = None,
         width: Optional[int] = None,
         buffer: Optional[NumType] = None,
+        all_touched: Optional[bool] = False,
+        apply_cutline: Optional[bool] = True,
```
vincentsarago, line 505: 🤔 This adds a bit of complexity because now you have to know if your data is of multiple resolution or not 🤷 

IMO for `feature` we should always have the mask representing the combination of the internal nodata/alpha and the given cutline geometry. 

I think we could handle this specific case within the `ImageData.create_from_list` method by rescaling the cutline mask and merging all mask in a 1 band layer that should save the maximum extent of all the masks

yellowcap: You are absolutely right! Just pushed a solution that works and even reduced the number of lines overall 🤓

yellowcap: Just saw that after running black on it the line count increased ➕ , but its still a way simpler solution I think

rio_tiler/models.py
```diff
@@ -421,18 +422,26 @@ def create_from_list(cls, data: Sequence["ImageData"]) -> "ImageData":
...
                 UserWarning,
             )
-            max_h, max_w = max(h), max(w)
             for img in data:
                 if img.height == max_h and img.width == max_w:
                     continue
                 arr = numpy.ma.MaskedArray(
                     resize_array(img.array.data, max_h, max_w),
-                    mask=resize_array(img.array.mask * 1, max_h, max_w).astype("bool"),
+                    mask=cutline_mask,
```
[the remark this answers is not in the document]
yellowcap: You are absolutely right again! But that opens the problem that pixels might never get unmasked when upsampling. In the following example, the green pixel would always be masked at the coarse resolution, but at the higher resolution it would not be masked. So the finish early would not work again. The solution is to go back to what you already suggested earlier - setting always  `all_touched` to `True`. In that case the finer mask should always be included in the coarser mask. Will make change accordingly.

image

rio_tiler/io/rasterio.py
```diff
@@ -551,7 +549,7 @@ def feature(
             [shape],
             out_shape=(img.height, img.width),
             transform=img.transform,
-            all_touched=all_touched,
+            all_touched=True,  # Necesary for matching masks at different resolutions
```
vincentsarago, line 552: 👍

yellowcap:
ok test for `multires` added, took the `blue.tif` test file, resampled from 900m to 1800m, and renamed the resampled file to `lowres.tif`. Added this as an asset to the `stac.json` item file, and using the lowres file in the test.

rio_tiler/io/rasterio.py
```diff
@@ -545,6 +545,20 @@ def feature(
             **kwargs,
         )
 
+        cutline_mask = rasterize(
+            [shape],
```
vincentsarago, line 549: @yellowcap I think there is something we are missing here. 

shape is assumed to be in the same CRS as the `img` crs but this is only the case when the user pass `dst_crs` equal `shape_crs` or `dst_crs=None` (because we always default to `shape_crs`.

before creating the `cutline_mask` I think you'll need to reproject the shape to the `img.crs`

we should also had better test because I feel this should have been caught (by older test)

rio_tiler/io/rasterio.py
```diff
@@ -521,17 +521,17 @@ def feature(
...
         """
+        shape = _validate_shape_input(shape)
+
         if not dst_crs:
             dst_crs = shape_crs
 
         # Get BBOX of the polygon
         bbox = featureBounds(shape)
 
-        cutline = create_cutline(self.dataset, shape, geometry_crs=shape_crs)
```
vincentsarago, line 530: we should also make sure that `img.statistics()` return the same result with and without this. 

without the cutline handled in GDAL we now rely only on the masked array. (I think it should be fine, but better test anyway)

vincentsarago: There are some difference, so I'll need to check the `ImageData.statistics` method
Screenshot 2023-05-02 at 11 32 54 AM

vincentsarago:
```python
# Using New feature method 
with Reader(src_path) as f:
    img_f = f.feature(feat)
    
# Using GDAL cutline 
with Reader(src_path) as f:
    cutline = create_cutline(f.dataset, feat, geometry_crs="epsg:4326")
    bbox = featureBounds(feat)
    img = f.part(bbox, vrt_options={"cutline": cutline})  

numpy.testing.assert_array_equal(img.mask, img_f.mask)
```

adding this kind of test could be nice (if not there yet), just to make sure we do not break anything.
The `img.data` won't be equal with the new version because we rely only on the masked array

rio_tiler/io/rasterio.py
vincentsarago:
```python
# Using New feature method 
with Reader(src_path) as f:
    img_f = f.feature(feat, dst_crs="epsg:3857")
    
# Using GDAL cutline 
with Reader(src_path) as f:
    cutline = create_cutline(f.dataset, feat, geometry_crs="epsg:4326")
    bbox = featureBounds(feat)
    img = f.part(bbox, vrt_options={"cutline": cutline}, dst_crs="epsg:3857")  

numpy.testing.assert_array_equal(img.mask, img_f.mask)
```

taken from https://github.com/cogeotiff/rio-tiler/pull/598/files#r1182322351

we can confirm this `bug` with ☝️

vincentsarago (changes requested):
We need to handle projection difference between the input feature and output crs

rio_tiler/io/rasterio.py
vincentsarago: no worries 😄

rio_tiler/io/rasterio.py
yellowcap: I would not expect this to be exactly the same, unless there is not projection going on. So if the dataset has the same projection as the feature and the `dst_crs` is not changed. Otherwise there might be differences from rasterizing the geometry in the projection of the dataset (vrt cutline) vs in the dst_crs (rasterize).

vincentsarago: Turns out there is a slight difference between using GDAL cutline and creating the mask using rasterize. It's definitely more `precise` with the rasterized mask (I guess this is due to reprojection of the cutline within warped vrt)




The `img.statistics()` will give different results but that's fine 👍

vincentsarago (approved):

vincentsarago:
thanks a ton @yellowcap for working on this 🙏

yellowcap:
Awesome! Thanks for all the help and getting it over  the last humps @vincentsarago !

PR #599, opened 2023-04-24
add flake8 to pre-commit and remove unused imports

yellowcap:
Used flake8 to remove unused imports. Also added flake8 to pre-commit hooks. But I suspect there could be a reason flake8 was not included already?


vincentsarago:
done in #600

Issue #605, opened 2023-05-15
update morecantile dependency to 4.0.0 (TMS 2.0)


vincentsarago:
https://github.com/developmentseed/morecantile/blob/main/CHANGES.md#400a0-2023-05-15

PR #606, opened 2023-05-15
allow morecantile 4.0

vincentsarago:
ref #605 

This PR changes how the TileMatrixSet are use to allow morecantile `3.1` and `>=4.0` version. The TMS value didn't change so in theory tiles created with Morecantile==3.1 and ==4.0 should be the same

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "httpx",
     "numexpr",
     "numpy",
-    "morecantile>=3.1,<4.0",
+    "morecantile>=3.1,<5.0",
     "pydantic",
     "pystac>=0.5.4",
     "rasterio>=1.3.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "httpx",
     "numexpr",
     "numpy",
-    "morecantile>=3.1,<4.0",
+    "morecantile>=3.1,<5.0",
```
vincentsarago, line 28: for now only `4.0.0a0` has been released so the CI will still use morecantile==3.3

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "httpx",
     "numexpr",
     "numpy",
-    "morecantile>=3.1,<4.0",
+    "morecantile>=4.0,<5.0",
```
vincentsarago, line 28: IMO it's safer to restrict the lower version to 4.0.
application using titiler might (like titiler 🙈) may not have pinned morecantile to specific versions

Issue #571, opened 2023-01-30
Use stac `Raster` extension to set the statistics 


vincentsarago:
from https://github.com/stac-extensions/raster#statistics-object

It will be a bit more complex because we need to add the statistics in the ImageModel after the `read` part which can have indexes or expression: https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/base.py#L478

Todo: 
- change https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/types.py#L30-L34 to add `stats` from the stac extension
- create the stats object and add it to the model after https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/base.py#L478

PR #611, opened 2023-05-26
forward statistics from STAC raster:bands

vincentsarago:
ref #571 

This PR Does:

- extend `AssetInfo` with `metadata` and `dataset_statistics` 
- forward the `statistics` within the ImageData (when not provided by the image itself)

## To Do

- [x] The `metadata` is not forwarded to the ImageData object when we create it from a list of ImageData https://github.com/cogeotiff/rio-tiler/blob/d0f9d11894f3b8a5d2e02fc637393873ddce0a57/rio_tiler/models.py#L458-L485


rio_tiler/reader.py
```diff
@@ -550,4 +552,5 @@ def point(
         coordinates=coordinates,
         crs=coord_crs,
         band_names=img.band_names,
+        metadata=dataset.tags(),
```
vincentsarago, line 555: by default we will forward the dataset metadata `tags`

rio_tiler/models.py
```diff
@@ -474,6 +481,12 @@ def create_from_list(cls, data: Sequence["ImageData"]) -> "ImageData":
         )
         dataset_statistics = stats if len(stats) == len(band_names) else None
 
+        metadata = dict(
+            itertools.chain.from_iterable(
+                [img.metadata.items() for img in data if img.metadata]
+            )
+        )
```
vincentsarago, line 488: we merge the metadata all together.
It's assumed that for MultiBands/MultiAssets the metadata will be in form of `{"asset1": {...}}`

Issue #603, opened 2023-04-26
Mask nodata values in XarrayReader's tile method


yoninachmany:
Users should be able to mask nodata values with Xarray datasets like they do with GeoTIFF datasets

References:
- https://github.com/cogeotiff/rio-tiler/pull/530#issuecomment-1523853038
- https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/xarray.py#L239
- https://corteva.github.io/rioxarray/html/getting_started/nodata_management.html

PR #612, opened 2023-05-30
handle nodata in XarrayReader

vincentsarago:
ref #603 

@yoninachmany it would be great if you could try this branch and see if it fits your request 🙏


vincentsarago:
Note: there is something going on with our CI. tests fails on the first attempt but then succeed 🤷

PR #628, opened 2023-07-18
Ensure that `dataset` is still open when reading `tags`

JackDunnNZ:
Previously this statement was outside the `ctx` context, and so `dataset` may already have been freed before we call `dataset.tags()`. In my experience, this was sporadically leading to either of the following errors:

```
rasterio.errors.RasterioIOError: Dataset is closed`
```

or 

```
rasterio._err.ObjectNullError: Pointer 'hObject' is NULL in 'GDALGetMetadata'.
```

The corresponding statement for the `read` method was already inside the `ctx` context, so this was only a problem for the `point` method.

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -547,10 +547,10 @@ def point(
             post_process=post_process,
         )
 
-    return PointData(
-        img.array[:, 0, 0],
-        coordinates=coordinates,
-        crs=coord_crs,
-        band_names=img.band_names,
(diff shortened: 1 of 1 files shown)


vincentsarago:
good catch 🙏

vincentsarago:
@JackDunnNZ I'm going to merge this and make a new release today 🙏

vincentsarago (approved):

JackDunnNZ:
Thanks!

PR #630, opened 2023-07-24
update morecantile and pydantic requirements

vincentsarago:
This PR does:
- update morecantile dependency to >=5.0,<6.0
- update models for pydantic 2.0
- update docs

## 5.1 or 6.0

There is no real API change between 5.0 and this PR, but the pydantic update is quite big. I wonder if we should do a major update?


vincentsarago:
After talking on slack: Let's do 6.0 🥳

sikmir:
I wonder if rio-tiler really is not compatible with morecantile 6.0?

vincentsarago:
@sikmir i'm not sure to understand your question

sikmir:
> @sikmir i'm not sure to understand your question

I'm about `morecantile>=5.0,<6.0`, does it really mean that rio-tiler doesn't expected to work with morecantile 6.0? or just maybe work, but untested?

vincentsarago:
The next version of rio-tiler (release on Monday) will work with morecantile 6.0 

https://github.com/cogeotiff/rio-tiler/blob/0ee8dfb06d703c6269de3ab26a71e633231e1e28/CHANGES.md#L80
https://github.com/cogeotiff/rio-tiler/blob/0ee8dfb06d703c6269de3ab26a71e633231e1e28/pyproject.toml#L29
https://github.com/cogeotiff/rio-tiler/pull/748

PR #635, opened 2023-08-07
Update `data_as_image` to return masked values

JackDunnNZ:
Prior to this change, the raw data array was being returned, which could include `nodata` values, throwing off an automatic image color scale. In the following example, the nodata value used is -9999 (for all oceans), which causes the image to basically appear as a land mask:

```
from matplotlib.pyplot import imshow
from rio_tiler.io.rasterio import Reader

with Reader("https://data.chc.ucsb.edu/products/CHIRPS-2.0/global_daily/cogs/p05/2023/chirps-v2.0.2023.02.01.cog") as r:
    img = r.read(nodata=-9999)
    
imshow(img.data_as_image())
```
image

If instead we use the masked array values, the automatic color scale is scoped to just the valid data values:

```
imshow(img.array[0, :, :])
```
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# 6.0.2 (unreleased)
+
+* Update `data_as_image` to return masked values (author @JackDunnNZ, https://github.com/cogeotiff/rio-tiler/pull/635)
 
 # 6.0.1 (2023-07-25)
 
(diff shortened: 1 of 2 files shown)


vincentsarago:
@JackDunnNZ will `imshow(img.data_as_image())` still work event if the return type is a Numpy Masked array?

JackDunnNZ:
Yes it seems to, the second example above is passing a masked array to `imshow`. I also get the same result running `imshow(img.data_as_image())` on this branch

vincentsarago:
@JackDunnNZ can you add something in the changelog 🙏

vincentsarago:
🙏

PR #637, opened 2023-09-13
Return a 1x1 image when bbox is smaller than a single pixel

JackDunnNZ:
This was originally motivated by hitting a zero-division error when calling `Reader.feature` with a feature that happened to be smaller than one pixel in the image:

```
File "site-packages/rasterio/transform.py", line 189, in from_bounds
    (east - west) / width, (south - north) / height)
    ~~~~~~~~~~~~^~~~~~~~~
ZeroDivisionError: float division by zero
```

I traced it back to the result of `part` being a 0x0 image when the bbox was smaller than a single pixel of the image, which this change fixes and returns a 1x1 image instead.

I'm not 100% sure about this, but it seems to make sense to me that `part` returning a single pixel is the correct outcome for this case, as this is still the valid image data for the bbox being requested.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+
+# 6.0.3 (2023-09-13)
+
+* return a 1x1 image when bbox is smaller than a single pixel (author @JackDunnNZ, https://github.com/cogeotiff/rio-tiler/pull/637)
+
 # 6.0.2 (2023-08-21)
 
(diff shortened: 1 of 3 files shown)


vincentsarago (approved):
🙏 that totally makes senses 

thank you

JackDunnNZ:
Thanks!

PR #640, opened 2023-09-20
Compute area-weighted statistics

vincentsarago:
ref https://github.com/NASA-IMPACT/veda-backend/issues/226

This PR adds a `coverage` options to the `get_array_statistics` function to enable `area weighted` statistics. This PR do not take care of the `coverage` array creation which should be done by the client application (maybe with some helper in rio-tiler)

cc @kylebarron @j08lue

### To Do

- [x] more tests
- [x] docs
- [x] validate the logic (why some values are defined using weight or not)
- [x] add `coverage` utility functions


rio_tiler/utils.py
```diff
@@ -36,6 +36,7 @@ def get_array_statistics(
     categorical: bool = False,
     categories: Optional[List[float]] = None,
     percentiles: Optional[List[int]] = None,
+    coverage: Optional[numpy.ndarray] = None,
```
kylebarron, line 39: Since I like really like typing 🙈 maybe
```py
from numpy.typing import NDArray
...
coverage: Optional[NDArray[np.floating]] = None
```
I think from reading this, it's expected to be a float? Even if it doesn't impact type checking much, I always find it helpful to recall what data types of arrays are expected where

rio_tiler/utils.py
```diff
@@ -121,21 +122,36 @@ def get_array_statistics(
...
-                "sum": float(data[b].sum()),
-                "std": float(data[b].std()),
-                "median": float(numpy.ma.median(data[b])),
-                "majority": float(keys[counts.tolist().index(counts.max())].tolist())
-                if valid_pixels
-                else numpy.nan,
-                "minority": float(keys[counts.tolist().index(counts.min())].tolist())
-                if valid_pixels
-                else numpy.nan,
+                "mean": float(array.mean()),
```
vincentsarago, line 148: we don't use the weighted array for the `min` and `max` (like https://github.com/isciences/exactextract#supported-statistics)

rio_tiler/utils.py
```diff
@@ -121,21 +122,36 @@ def get_array_statistics(
...
+        else:
+            array = data[b]
+
+        count = array.count() if coverage is None else coverage.sum()
+        if valid_pixels:
+            majority = float(keys[counts.tolist().index(counts.max())].tolist())
+            minority = float(keys[counts.tolist().index(counts.min())].tolist())
+        else:
+            majority = numpy.nan
+            minority = numpy.nan
```
vincentsarago, line 142: This is unrelated changes for styling

rio_tiler/utils.py
```diff
@@ -121,21 +122,36 @@ def get_array_statistics(
...
-                else numpy.nan,
+                "mean": float(array.mean()),
+                "count": float(count),
+                "sum": float(array.sum()),
+                "std": float(array.std()),
+                "median": float(numpy.ma.median(array)),
+                "majority": majority,
+                "minority": minority,
                 "unique": float(counts.size),
                 **dict(zip(percentiles_names, percentiles_values)),
```
vincentsarago, line 156: 🤔 maybe the percentiles should use the weighted array?

tests/test_utils.py
```diff
@@ -561,3 +561,30 @@ def parse(content):
...
+
+
+def test_get_array_statistics_coverage():
+    """Test statistics with coverage array."""
+    # Data Array
+    # 1, 2
+    # 3, 4
+    data = np.ma.array((1, 2, 3, 4)).reshape((1, 2, 2))
+
+    # Coverage Array
```
kylebarron, line 573: Maybe we should call this the `weight` array instead of `coverage`? `coverage` doesn't seem to make the most sense to me here.

kylebarron (approved):

tests/test_utils.py
vincentsarago: in https://github.com/isciences/exactextract, weight are really weight, while the `coverage` is called `cell coverage fractions`. I didn't want to confuse people 🤷

kylebarron: I see, `coverage` came from exactextract because their weights are usually derived from the coverage of each polygon in the cell. I think `weight` is more general than `coverage` though. There could be use cases for weighted zonal stats that aren't partial-coverage

vincentsarago: `coverage_weights` ?

kylebarron: Would we ever have both `coverage` and `weights`?

kylebarron: It looks like exactextract does support both `coverage` and `weight` in effect, because it allows `weight` as a parameter and it computes `coverage` itself

vincentsarago: > Would we ever have both coverage and weights?

Not planned right now but I don't want to be blocked in the future. I agree that `coverage` is not a perfect name but I don't want to use `weights` because it's not it's not `weights` but spatial fraction. I'm open to change to better name if you have ideas :D

rio_tiler/models.py
```diff
@@ -787,3 +790,43 @@ def statistics(
...
+            transform=self.transform * Affine.scale(1 / cover_scale),
+            all_touched=True,
+            fill=0,
+            dtype="uint8",
+        )
+        cover_array = cover_array.reshape(
+            (self.height, cover_scale, self.width, cover_scale)
+        ).astype("float32")
+
+        return cover_array.sum(-1).sum(1) / (cover_scale**2)
```
vincentsarago, line 832: slightly modified version of https://github.com/perrygeo/python-rasterstats/pull/136

@sgoodm I know ☝️ is a 7 years old PR (😅) but I hope you don't mind that I reused some of the code here 🙏

vincentsarago:
Note: we don't have a BaseReader method that returns statistics for a GeoJSON Feature but this is how it could look like in the `client` (e.g TiTiler):

```python
with Reader(path) as src:
    data = src_dst.feature(
        shape,
        shape_crs=WGS84_CRS,
    )

    coverage_array = data.get_coverage_array(
        shape, shape_crs=WGS84_CRS
    )

    stats = data.statistics(coverage=coverage_array)

```

rio_tiler/models.py
sgoodm: Not at all, @jacobwhall and I have been pushing use of COGs in our work at @aiddata and are fans of these projects, so happy to share some code to support :+1: :+1:

kylebarron (approved):

PR #651, opened 2023-11-02
Proof of concept to allow iterable instead of sequence for mosaics.

bitner:
Pystac-Client returns an iterable that uses the paging mechanism on search results. 

By allowing an iterable rather than a sequence, we can implement titiler-pgstac like behavior on any stac api.

Basically, the max_items would be comparable to the _scan_limit parameter used in titiler-pgstac.

In this example, it will keep scanning through the iterable until it either hits the max_items or until pixel_selection is done.

The iterable_example.ipynb notebook demos this for a single tile.

### DIFF
--- a/rio_tiler/mosaic/reader.py
+++ b/rio_tiler/mosaic/reader.py
@@ -2,7 +2,7 @@
 
 import warnings
 from inspect import isclass
-from typing import Any, Callable, List, Optional, Sequence, Tuple, Type, Union, cast
+from typing import Any, Callable, List, Optional, Sequence, Tuple, Type, Union, cast, Iterable
 
 import numpy
 from rasterio.crs import CRS
@@ -23,7 +23,7 @@ from rio_tiler.utils import _chunks, resize_array
 
 
 def mosaic_reader(  # noqa: C901
-    mosaic_assets: Sequence,
+    mosaic_assets: Iterable,
     reader: Callable[..., ImageData],
     *args: Any,
     pixel_selection: Union[Type[MosaicMethodBase], MosaicMethodBase] = FirstMethod,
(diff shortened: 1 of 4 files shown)


rio_tiler/mosaic/reader.py
```diff
@@ -23,7 +23,7 @@
 
 
 def mosaic_reader(  # noqa: C901
-    mosaic_assets: Sequence,
+    mosaic_assets: Iterable,
```
vincentsarago, line 26: I think we could support both Iterable and Sequence so we don't break things

rio_tiler/mosaic/reader.py
```diff
@@ -76,15 +76,17 @@ def mosaic_reader(  # noqa: C901
             "'rio_tiler.mosaic.methods.base.MosaicMethodBase'"
         )
 
-    if not chunk_size:
-        chunk_size = threads if threads > 1 else len(mosaic_assets)
+    # if not chunk_size:
+    #    chunk_size = threads if threads > 1 else len(mosaic_assets)
```
vincentsarago, line 80: The only downside of Iterable is that we don't know its length but I'm fine not checking this for Iterable!

rio_tiler/mosaic/reader.py
```diff
@@ -76,15 +76,17 @@ def mosaic_reader(  # noqa: C901
...
+    chunk_size = threads
 
     assets_used: List = []
     crs: Optional[CRS]
     bounds: Optional[BBox]
     band_names: List[str]
 
     for chunks in _chunks(mosaic_assets, chunk_size):
```
vincentsarago, line 88: Here we should translate a Sequence to Iterable so it will work in `_chunks()`

Issue #654, opened 2023-11-15
Investigate weird output resolution after #648


vincentsarago:
with rio-tiler `6.2.4`

<img width="1172" alt="Screenshot 2023-11-15 at 9 32 07 PM" src="https://github.com/cogeotiff/rio-tiler/assets/10407788/0b0bb5b1-59b1-4ec4-8a0f-76b4245ef4b2">

with rio-tiler `6.2.3`

<img width="1149" alt="Screenshot 2023-11-15 at 9 31 42 PM" src="https://github.com/cogeotiff/rio-tiler/assets/10407788/ac929d51-52df-4b81-8235-04affaf6d850">

data: s3://ghgc-data-store-dev/NASA_GSFC_ch4_wetlands_daily/NASA_GSFC_ch4_wl_ch4_wetlands_v22_x720_y360_daily_20211231.tif

Pretty sure this is due to the change in #648

vincentsarago:
Ok I think I know what's going on! 

The file has bounds: `BoundingBox(left=-180.0, bottom=-90.0, right=180.0, top=90.0)`, when we want tiles in WebMercator we call `rasterio.warp.calculate_default_transform` to get the transform of the `reprojected` input file... but our file bounds is much bigger than the webmercator bounds so the transform is :poop: 


https://github.com/cogeotiff/rio-tiler/blob/dc7e87d68b08f107005c6ee5273df0b5b119b657/rio_tiler/utils.py#L247-L249

files looks good in EPSG:4326 projection

<img width="500" alt="Screenshot 2023-11-15 at 10 22 56 PM" src="https://github.com/cogeotiff/rio-tiler/assets/10407788/bb43f5f4-bad5-48c7-a395-c283608ebe6b">

```python
import rasterio
from rasterio.warp import calculate_default_transform
from rasterio.warp import transform_bounds

bbox = (-160, -70, 160, 70)
print("bbox read: ", bbox)
b = transform_bounds("epsg:4326", "epsg:3857", *bbox)
print("bbox read (epsg: 3857): ", b)
w, s, e, n = b

print()
with rasterio.open("tests/fixtures/cog_nasa.tif") as src_dst:
    print("dataset bbox :", src_dst.bounds)
    dst_transform, _, _ = calculate_default_transform(
        src_dst.crs, "epsg:3857", src_dst.width, src_dst.height, *src_dst.bounds
    )
    print(w, h)
    print("res: ", dst_transform.a, dst_transform.e)
[...]

PR #660, opened 2023-11-29
adjust latitude in WarpedVRT parameter calculation

vincentsarago:
Simple fix for #654


rio_tiler/utils.py
```diff
@@ -244,9 +244,38 @@ def get_vrt_transform(
...
+        # of the WebMercator projection, we `crop` the dataset
+        # to get the transform (resolution).
+        #
+        # Note: Should be handled in gdal 3.8 directly
+        # https://github.com/OSGeo/gdal/pull/8775
+        if (
+            src_dst.crs == WGS84_CRS
+            and dst_crs == WEB_MERCATOR_CRS
+            and (src_bounds[1] < -85.06 or src_bounds[3] > 85.06)
+        ):
```
vincentsarago, line 264: In theory we should do this for all projection combination but rasterio does not surface CRS `area of use`

PR #670, opened 2024-01-16
adds `align_bounds_with_dataset` to adjust bounds in reader.part() method

vincentsarago:
Thanks to @abarciauskas-bgse for deep-diving into this issue.

When doing zonal statistics, users might not want to get the stats from resampled data but `clip` the data to the closest rounding window (full pixel) and then use the `coverage` option within the statistics method to adjust the values.

Screenshot 2024-01-16 at 12 38 42 PM

With this PR we're introduction an option in `rio_tiler.reader.part` to adjust the bounds to the dataset transform. This option might be used from `feature() or `part()` reader's method such as 

```python
with Reader("cog.tif") as src:
    data = src.feature(
        shape,
        shape_crs=WGS84_CRS,
        align_bounds_with_dataset=True,
    )

    coverage_array = data.get_coverage_array(
        shape,
        shape_crs=WGS84_CRS,
    )

    stats = data.statistics(coverage=coverage_array)
```

### DIFF
--- a/docs/src/advanced/statistics.md
+++ b/docs/src/advanced/statistics.md
@@ -1,5 +1,5 @@
 
-### Statistics
+#### Form `Readers`
 
 `rio-tiler`'s Readers provide simple `.statistics` method to retrieve dataset global statistics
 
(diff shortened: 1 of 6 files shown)


docs/src/advanced/statistics.md
```diff
@@ -133,34 +69,110 @@ assert stats[0]["mean"] == 1.125  # (1 * 0.5 + 2 * 0.0 + 3 * 1.0 + 4 * 0.25) / 4
...
+        shape_crs=WGS84_CRS,
     )
 
-    # Get statistics (ImageData.statistics is calling `rio_tiler.utils.get_array_statistics`)
     stats = data.statistics(coverage=coverage_array)
 ```
 
-!!! warnings
-    The coverage weights will only have influence on specific statistics:
+When passing `align_bounds_with_dataset=True` to the  `reader.part()` method (forwared from `.feature` or `.part` reader's method), rio-tiler will adjust the input geometry bounds to math the input dataset resolution/transform and avoid unnecessary resampling.
```
abarciauskas-bgse, line 92: When passing `align_bounds_with_dataset=True` to the `reader.part()` method (forwarded from `.feature` or `.part` reader methods), rio-tiler will adjust the input geometry bounds to match the input dataset resolution/transform and avoid unnecessary resampling.

rio_tiler/utils.py
```diff
@@ -299,6 +302,19 @@ def get_vrt_transform(
         # Get Bounds for the rounded window
         bounds = src_dst.window_bounds(w)
 
+    elif align_bounds_with_dataset:
```
abarciauskas-bgse, line 305: I think this code is the same as in reader.py - should it be a function in this file and re-used in reader.py?

PR #681, opened 2024-02-26
Fix weighted statistics

### LINKED ISSUE #680: Incorrect array statistics with (coverage) weights
[6.2.0](https://github.com/cogeotiff/rio-tiler/blob/main/CHANGES.md#620-2023-09-27) introduced the possibility to calculate array statistics with `coverage` weights.

However, we need to review the way the weighted stats are calculated, in the `get_array_statistics` function.

https://github.com/cogeotiff/rio-tiler/blob/73175d0607ef28bffcb6f03ff8dfd2d6abfbc191/rio_tiler/utils.py#L36

1. `mean` - we currently take the mean of the pixel values scaled by the weights. It should be the sum of scaled values divided by sum of weights - or use `numpy.average(..., weights=coverage)`.
   ```python
   assert stats[0]["mean"] == 1.125  # (1 * 0.5 + 2 * 0.0 + 3 * 1.0 + 4 * 0.25) / 4
   ```
   should be `/ 1.75`
2. `std`, `percentiles`, `median` - we need to use weighted stats for these. Those are simple Numpy formulae - [weighted standard deviation](https://stackoverflow.com/a/2415343), [weighted percentiles](https://stackoverflow.com/a/29677616), and the weighted median would be the 0.5 weighted percentile.
3. `count` - what is that for weighted data? Count of all finite values where weights are non-zero? Or the sum of the weights?
4. `min`, `max`, `minority`, `majority`, `unique` and also `histogram` - we should probably only take values into account where weights are non-zero.
5. `sum` is correct (sum of weighted values)

j08lue:
Fixes #680

- [x] Change the simple tests for `get_array_statistics` to reflect more details of weighted stats - tests will fail ❌
- [ ] Add tests for more weighted stats
- [ ] Implement correct weighted statistics in `get_array_stats`

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -567,21 +567,17 @@ def test_render_colorinterp():
 def test_get_array_statistics_coverage():
     """Test statistics with coverage array."""
     # Data Array
-    # 1, 2
-    # 3, 4
     data = np.ma.array((1, 2, 3, 4)).reshape((1, 2, 2))
 
     # Coverage Array
-    # 0.5, 0
-    # 1, 0.25
-    coverage = np.array((0.5, 0, 1, 0.25)).reshape((2, 2))
+    coverage = np.array((0, 0.25, 1, 0)).reshape((2, 2))
 
     stats = utils.get_array_statistics(data, coverage=coverage)
     assert len(stats) == 1
-    assert stats[0]["min"] == 1
-    assert stats[0]["max"] == 4
-    assert stats[0]["mean"] == 1.125  # (1 * 0.5 + 2 * 0.0 + 3 * 1.0 + 4 * 0.25) / 4
-    assert stats[0]["count"] == 1.75
+    assert stats[0]["min"] == 2
+    assert stats[0]["max"] == 3
+    assert stats[0]["mean"] == (1 * 0 + 2 * 0.25 + 3 * 1.0 + 4 * 0) / 1.25
+    assert stats[0]["count"] == 1.25
 
     stats = utils.get_array_statistics(data)
     assert len(stats) == 1


tests/test_utils.py
```diff
@@ -567,21 +567,17 @@ def parse(content):
...
     stats = utils.get_array_statistics(data, coverage=coverage)
     assert len(stats) == 1
-    assert stats[0]["min"] == 1
-    assert stats[0]["max"] == 4
-    assert stats[0]["mean"] == 1.125  # (1 * 0.5 + 2 * 0.0 + 3 * 1.0 + 4 * 0.25) / 4
-    assert stats[0]["count"] == 1.75
+    assert stats[0]["min"] == 2
+    assert stats[0]["max"] == 3
+    assert stats[0]["mean"] == (1 * 0 + 2 * 0.25 + 3 * 1.0 + 4 * 0) / 1.25
+    assert stats[0]["count"] == 1.25
```
vincentsarago, line 580: Let's look at the data 

```python
data = np.ma.array((1, 2, 3, 4)).reshape((1, 2, 2))
coverage = np.array((0, 0.25, 1, 0)).reshape((2, 2))
data * coverage

>> masked_array(
  data=[[[0. , 0.5],
         [3. , 0. ]]],
  mask=False,
  fill_value=1e+20)
```

the stats should then be:

```python
min: 0
max: 3
mean: (0 + 0.5 + 3.0 + 0.) / 4  = 0.875
sum: 0 + 0.5 + 3.0 + 0. = 3.5
count: 0 + 0.25 + 1 + 0 = 1.25 
```

I'm not sure to understand why the `mean` should be the sum of the `data * coverage` divided by the sum of the coverage. We already apply the coverage factor so we (IMO) just need to divide by the number of pixel

j08lue: Because `coverage` does not sum up to 1. Its size is arbitrary and not related to the overall weight of all cells, so it skews the overall quantity.

A simple case illustrates this: imagine 2 x 2 cells, all containing pixel value 20, the coverage is the same - 0.1 for all cells.

Since the coverage / weight is the same for all cells, you would expect that their weighted average is the same as their simple average, namely 20, right?

Simple average:
```
(20 + 20 + 20 + 20) / 4 = 20
```

Weighted sum:
```
(20 * 0.1 + 20 * 0.1 + 20 * 0.1 + 20 * 0.1) = 8
```

If you divide that by 4 (the number of cells), you get 2.


You need to divide by the sum of the weights to get the expected result:
```
8 / (0.1 + 0.1 + 0.1 + 0.1) = 20
```

This is not a proof, but perhaps still helps?

j08lue: https://en.wikipedia.org/wiki/Weighted_arithmetic_mean

$\bar{x} = \frac{w_1 x_1 + w_2 x_2 + ... + w_n x_n}{w_1 + w_2 + ... + w_n}$

vincentsarago: well I'm not sure to full understand but in our case we don't use `weight` but `% of each pixel, coverage` so the sum of all the weight should be the number of pixels not the sum of coverage

PR #688, opened 2024-03-22
Feat/set profile in imagedata

iferencik:
@vincentsarago 

This PR adds an optional **profile** attribute to ImageData  in case the reader that created it was requested to unscale the data.

as per https://github.com/developmentseed/titiler/discussions/803

The profile attr will contain the scales and offsets, which at this point are not part of the profile attribute of the rio DatasetReader

This is to accomodate a specific use case when the reader is not STAC or Mosaic.

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -283,7 +283,7 @@ def read(
             band_names=[f"b{idx}" for idx in indexes],
             dataset_statistics=dataset_statistics,
             metadata=dataset.tags(),
-            profile=dataset.profile
+            profile=dataset_profile
         )
 
 
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -256,11 +256,18 @@ def read(
         # TODO: DEPRECATED, masked array are already using bool
         if force_binary_mask:
             pass
-
+        # profile is added  only if unscaling was requested. This would ensure/signalized
+        # an advanced usage specifically applied to COGS
+        dataset_profile = None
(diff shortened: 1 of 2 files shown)


rio_tiler/reader.py
```diff
@@ -256,11 +256,18 @@ def read(
...
+        dataset_profile = None
         if unscale:
             data = data.astype("float32", casting="unsafe")
             numpy.multiply(data, dataset.scales[0], out=data, casting="unsafe")
             numpy.add(data, dataset.offsets[0], out=data, casting="unsafe")
+            dataset_profile = dataset.profile
+            if not 'scales' in dataset_profile:
+                dataset_profile['scales'] = dataset.scales
+            if not 'offsets' in dataset_profile:
+                dataset_profile['offsets'] = dataset.offsets
```
vincentsarago, line 270: what if we want the scales and offsets info even if we don't `apply` them?

vincentsarago: alos I think if we add `profile` we should always populate the field, not just when we `unscale=True`

iferencik: well, i made it conditional on unscale request as to avoid issues with STAC and Mosaic readers.
I believe the profile probably is always available but i did this per your suggestion.

It would be easy to remove the condition and update the profile with **"scales" and "offsets"**

vincentsarago:
this was done in https://github.com/cogeotiff/rio-tiler/pull/818

PR #695, opened 2024-04-08
Error with expression with unicode dtype

### LINKED ISSUE #694: Error applying expression with STACReader : "NumExpr 2 does not support Unicode as a dtype."
#### Problem description

On titiler, doing a request similar to `/stac/tiles/WebMercatorQuad/9/451/76.png?url=s3%3A%2F%2Fdc-development-catalog%2Fcalls%2Fcall-969%2FcalibratedDatasets%2FS2A_MSIL2A_20230928T042551_N0509_R133_T54XVM_20230928T053752-calibrated%2FS2A_MSIL2A_20230928T042551_N0509_R133_T54XVM_20230928T053752-calibrated.json&rescale=0%2C1&resampling_method=average&resampling=nearest&expression=where((s2cloudless_clp%3E0.5)%2C1%2C0)%3B&asset_as_band=true`

Titiler returns the following error:
```json
{
  "detail": "NumExpr 2 does not support Unicode as a dtype."
}
```

The error is raised when the `expression` parameter is set.

#### Expected Output

Correctly processed tile

#### Environment Information

Titiler from 0.14.0 to 0.18.0
container `ghcr.io/developmentseed/titiler:0.18.0` + `boto3` and `rio-stac`

#### Analysis

I could reproduce the error in rio-tiler in the following branch: https://github.com/emmanuelmathot/rio-tiler/tree/U17

emmanuelmathot:
This PR fixes #694.
It checks that the statistics provided in the STAC item for an asset are valid. If not, a warning is output.
@vincentsarago If you consider stats must be valid without warning, drop this PR.


tests/test_io_stac.py
```diff
@@ -848,3 +849,38 @@ def test_metadata_from_stac(rio):
...
+        
+        asset_info = stac._get_asset_info("s2cloudless")
+        url = asset_info["url"]
+        with stac.reader(url, tms=stac.tms, **stac.reader_options) as src:
+            img = src.tile(451, 76, 9, expression='where((b1>0.5),1,0)')
+            assert img.data.shape == (1, 256, 256)
+            assert img.mask.shape == (256, 256)
+            assert img.band_names == ["where((b1>0.5),1,0)"]
+        
+        img = stac.tile(451, 76, 9, expression='where((s2cloudless>0.5),1,0)', asset_as_band=True)
```
emmanuelmathot, line 883: It raises the issue here whilst the same expression directly on the COG reader before worked properly

emmanuelmathot:
error found. Statistics in the stac asset are strings. It is not the first time we encounter this issue.
@vincentsarago Should we make a check of the statistics somewhere in STACReader and output a warning in case of non number values?

vincentsarago:
@emmanuelmathot 🤯  thanks for looking at this 🙏 

is it in the specification that the stats could be string? If yes we can add a fix in https://github.com/cogeotiff/rio-tiler/blob/89fac81e011218f64071857a70244976809a48c7/rio_tiler/io/stac.py#L303-L308

emmanuelmathot:
no it must be a double, this is an issue on our side but it is a common mistake and any warning clue on a error like this is welcome

vincentsarago:
@emmanuelmathot Sadly rio-tiler cannot account for all error that could happen in STAC 😬 

If we start adding patch for some things I'm afraid that we would receive PR for others 🤷

emmanuelmathot:
I totally understand. My idea was just to check the data in `update_statistics` method and make a warning if the stac data are not valid but if you think it will create overheads in the code, let's drop it.

rio_tiler/io/stac.py
```diff
@@ -306,6 +307,14 @@ def _get_asset_info(self, asset: str) -> AssetInfo:
...
                 if {"minimum", "maximum"}.issubset(b.get("statistics", {}))
             ]
+            # check that stats data are all double and make warning if not
+            if stats and not all(
+                isinstance(v, (int, float)) for stat in stats for v in stat
+            ):
+                warnings.warn(
+                    "Some statistics data are not double, this may cause issues in data processing."
+                )
+
```
vincentsarago, line 317: I'm not sure it makes 💯 sense just to raise a warning because this will likely create an Error later. 

I think we could raise a warning AND not add the `invalid` statistics.

emmanuelmathot: sure. that is even better

vincentsarago (approved):

PR #696, opened 2024-04-17
make sure rio-tiler works with files with Inverted Lat (transform)

vincentsarago:
ref: https://github.com/US-GHG-Center/veda-config-ghg/pull/333


vincentsarago:
something changed with rasterio 1.3.10, I will need to debug before being able to merge this PR

PR #701, opened 2024-05-02
use width/height in `get_vrt_transform` when overzooming a dataset

vincentsarago:
This PR re-add `width` and `height` parameters to the `get_vrt_transform` (removed in #648) but will only be used when over-zooming a dataset

cc @amarouane-ABDELHAK @j08lue

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -296,8 +296,8 @@ def get_vrt_transform(
     Args:
         src_dst (rasterio.io.DatasetReader or rasterio.io.DatasetWriter or rasterio.vrt.WarpedVRT): Rasterio dataset.
         bounds (tuple): Bounding box coordinates in target crs (**dst_crs**).
-        height (int, optional): Desired output height of the array for the input bounds.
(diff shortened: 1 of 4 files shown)


rio_tiler/utils.py
```diff
@@ -384,6 +384,13 @@ def get_vrt_transform(
 
     w, s, e, n = bounds
 
+    if height and width:
+        out_transform = from_bounds(w, s, e, n, width, height)
+        # when not overzooming we don't want to set width/height output
+        if (out_transform.a > dst_transform.a) and (out_transform.e < dst_transform.e):
```
j08lue, line 390: This seems to assume that `e` is negative - may need to `abs` the pixel sizes for a safe comparison?

amarouane-ABDELHAK:
Thank you for offering a permanent solution to the issue. It was indeed a challenge to identify the cause of the visualization problem. I wonder why we didn't think of integrating these additional lines of code into the existing [pull request](https://github.com/cogeotiff/rio-tiler/pull/700) rather than starting a new one.

PR #707, opened 2024-05-15
Update unscaling to use the per-band scale/offset values

jddeal:
Hey there!

This is a small change to extend `unscale=True` to lean on the per-band values instead of assuming the first scale/offset applies to all bands.

(Sponsored by [Vaisala Xweather](https://www.xweather.com/))


vincentsarago:
@jddeal I'm not quite sure why we need this. In GDAL/Rasterio I thought the metadata model was to have only one `scale/offset` value 🤷‍♂️

tests/test_reader.py
```diff
@@ -429,8 +429,8 @@ def test_point():
         assert pt.band_names == ["b1"]
 
         pt = reader.point(src_dst, [310000, 4100000], coord_crs=src_dst.crs)
-        assert pt.data == numpy.array([8917])
-        assert pt.band_names == ["b1"]
+        numpy.testing.assert_equal(pt.data, [8917, 1001])
+        assert pt.band_names == ["b1", "b2"]
```
vincentsarago, line 433: 🤔 why do we get a 2 bands dataset when the input COG as only one band?

jddeal:
I suspect we might have a rather uncommon use case, but we haven't run into any limitations through gdal or rasterio that would imply only one scale/offset should be set per file.

Some example docs:
- [gdal_edit.py](https://gdal.org/programs/gdal_edit.html#cmdoption-scale) will set either one value for all bands or per-band values
- [rasterio DatasetReader](https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.DatasetReader.scales) returns per-band values

tests/test_reader.py
jddeal: I adjusted the fixture COG (tests/fixtures/cog_scale.tif) to include a second band with the unscaled values and no scale/offset metadata set. In the call above it doesn't pass `indexes=1` like the test above, so it uses both bands in the file.

four43:
> @jddeal I'm not quite sure why we need this. In GDAL/Rasterio I thought the metadata model was to have only one `scale/offset` value 🤷‍♂️

I will be forever sad that the nodata value is applied per file, not per band anymore. That's such a shame for heterogeneous data :(

vincentsarago:
@jddeal I don't seems to have push right on the fork so I cannot edit this PR

rio_tiler/reader.py
```diff
@@ -261,8 +261,13 @@ def read(
 
         if unscale:
             data = data.astype("float32", casting="unsafe")
-            numpy.multiply(data, dataset.scales[0], out=data, casting="unsafe")
-            numpy.add(data, dataset.offsets[0], out=data, casting="unsafe")
+
+            # reshaped to match data
+            scales = numpy.array(dataset.scales).reshape((-1,) + (1,) * (len(data.shape[1:])))
+            offsets = numpy.array(dataset.offsets).reshape((-1,) + (1,) * (len(data.shape[1:])))
```
vincentsarago, line 267:
```suggestion
            scales = numpy.array(dataset.scales).reshape(-1, 1, 1)
            offsets = numpy.array(dataset.offsets).reshape(-1, 1, 1)
```
should be enough, I don't see rasterio handling non 2D data

vincentsarago:
edited in https://github.com/cogeotiff/rio-tiler/pull/708

PR #709, opened 2024-05-17
Per-band unscaling works with indexes

jddeal:
This updates the per-band unscaling to support when indexes=[...] are passed. I didn't account for that in the PR a couple days ago, sorry about that!

Fortunately a quick fix to select from the scales/offsets arrays so that they continue to match the shape of the image data. I've been testing this with files we make containing dozens of bands with different (or no) scale/offset rules and I haven't noticed any issues with mismatching or incorrect shapes.

You also should be able to edit this forked repo now. Not sure what I'm missing in the settings, but it is part of my work org and maybe there's something I can't see getting in the way by default. I added you with write permission on the repo to hopefully get around that.

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -263,8 +263,8 @@ def read(
             data = data.astype("float32", casting="unsafe")
 
             # reshaped to match data
-            scales = numpy.array(dataset.scales).reshape(-1, 1, 1)
-            offsets = numpy.array(dataset.offsets).reshape(-1, 1, 1)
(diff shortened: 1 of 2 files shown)


vincentsarago:
oohhh thanks for catching this 😅 🙏

vincentsarago:
BTW @jddeal I don't need to be added to `aerisweather/rio-tiler` repo. Usually when creating a PR from fork you can allow maintainers to push directly in the fork https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

vincentsarago:
finished in #710

PR #712, opened 2024-06-05
Error message should match x, y, z arguments

abarciauskas-bgse:
It's a bit confusing when the error message has a different order than the arguments, so this is a small change to the error message.

### DIFF
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -232,7 +232,7 @@ class XarrayReader(BaseReader):
             reproject_method = resampling_method
 
         if not self.tile_exists(tile_x, tile_y, tile_z):
-            raise TileOutsideBounds(f"Tile {tile_x}/{tile_y}/{tile_z} is outside bounds")
+            raise TileOutsideBounds(f"Tile(x={tile_x}, y={tile_y}, z={tile_z}) is outside bounds")
 
         ds = self.input
         if nodata is not None:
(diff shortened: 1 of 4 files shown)


rio_tiler/io/xarray.py
```diff
@@ -232,7 +232,7 @@ def tile(
             reproject_method = resampling_method
 
         if not self.tile_exists(tile_x, tile_y, tile_z):
-            raise TileOutsideBounds(f"Tile {tile_z}/{tile_x}/{tile_y} is outside bounds")
+            raise TileOutsideBounds(f"Tile {tile_x}/{tile_y}/{tile_z} is outside bounds")
```
vincentsarago, line 235: maybe we should print something like `Tile(x={tile_x}, y={tile_y}, z={tile_z})`

vincentsarago: would you also be able to update https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/rasterio.py#L344-L347, https://github.com/cogeotiff/rio-tiler/blob/91095a27e663c76f2dcfdac64d4364ad9f9b9585/rio_tiler/io/base.py#L1037-L1040 and https://github.com/cogeotiff/rio-tiler/blob/91095a27e663c76f2dcfdac64d4364ad9f9b9585/rio_tiler/io/base.py#L481-L484 🙏

vincentsarago:
🙏 thanks @abarciauskas-bgse

abarciauskas-bgse:
Thanks @vincentsarago for making the additional changes and merging. I was going to write a test for the error message but 👌🏽 with me to forego that for now.

Issue #715, opened 2024-07-26
reader.point assumes latitude coordinate is increasing


georgespill:
`reader.point` checks whether coordinates are within the dataset bounds, but assumes that `bounds[1] < bounds[3]`, which is not the case for inverted latitude coordinates

PR #716, opened 2024-07-26
[715] Enable handling of inverted latitude in reader.point

georgespill:
`reader.point` raises a `PointOutsideBounds` error if the input coordinates are outside the dataset bounds. However, the current implementation assumes that `bounds[1] < bounds[3]`. This is not true for inverted latitude coordinates. This PR updates the implementation to handle the inverted case.

Fixes https://github.com/cogeotiff/rio-tiler/issues/715

See https://github.com/cogeotiff/rio-tiler/pull/696 for a similar issue

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # unreleased
 
 * better error message for `TileOutsideBounds` errors (author @abarciauskas-bgse, https://github.com/cogeotiff/rio-tiler/pull/712)
+* handle of inverted latitude in `reader.point` (author @georgespill, https://github.com/cogeotiff/rio-tiler/pull/716)
 
 # 6.6.1 (2024-05-17)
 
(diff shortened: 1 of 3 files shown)


vincentsarago:
thanks @georgespill 

do you thing we should had the same kind of warning as in https://github.com/cogeotiff/rio-tiler/pull/696/files#diff-aee02035a0c41599c6860bf04d8075994bb3f9e5b1d5d0595c0d02c5f4ddf596R50-R53 ?

would you be able to write a quick tests for this 🙏

georgespill:
Yep, adding the warning is a good suggestion. Have added that along with updating tests

georgespill:
@vincentsarago would it be possible to publish a release with this fix included?

vincentsarago:
@georgespill yes, I'm just waiting on https://github.com/cogeotiff/rio-tiler/pull/726 to get merged 🙏

PR #722, opened 2024-08-23
Adding `default_assets` for MultiBaseReader and STACReader

mccarthyryanc:
I have a specific use case when using `titiler` where I can't pass `assets` to the `STACReader` methods. As a solution to this, I added an optional `default_assets` attribute to be used like:
```python
with STACReader(STAC_PATH, default_assets=["green"]) as stac:
    img = stac.tile(71, 102, 8)
```
PR includes tests and updates to docs.

ref: https://github.com/cogeotiff/rio-tiler/discussions/720


vincentsarago:
@mccarthyryanc I'll try to review/merge this PR this week. 

Quick question, are you planning to use this with TiTiler? right now this won't work without customizing the Dependencies because the default dependencies expect assets/expression to be set or it will raise a 500 errors. At runtime the dependencies can check if the item as `default_assets` set so this will be pretty hard to enable `dynamic` check. 

Note: this could be resolved using the Render extension at a collection level

rio_tiler/io/base.py
```diff
@@ -495,9 +497,11 @@ def tile(
         if expression:
             assets = self.parse_expression(expression, asset_as_band=asset_as_band)
 
+        assets = assets or self.default_assets
```
vincentsarago, line 500: what do you think about raising a Warning here when we use the default_assets? 

```python
if not assets and self.default_assets:
    warnings.warn(
        f"No assets/expression passed, defaults to {self.default_assets}", UserWarning
    )
    assets = self.default_assets
```

It adds couple lines 🤷‍♂️

rio_tiler/io/stac.py
```diff
@@ -200,6 +200,7 @@ class STACReader(MultiBaseReader):
         exclude_assets (set of string, optional): Exclude specific assets.
         include_asset_types (set of string, optional): Only include some assets base on their type.
         exclude_asset_types (set of string, optional): Exclude some assets base on their type.
+        default_assets (set of string, optional): Default assets to use if none are defined.
```
vincentsarago, line 203:
```suggestion
        default_assets (list of string, optional): Default assets to use if none are defined.
```

rio_tiler/io/base.py
mccarthyryanc: sounds good to me!

mccarthyryanc:
Thanks for the review!

> Quick question, are you planning to use this with TiTiler? right now this won't work without customizing the Dependencies because the default dependencies expect assets/expression to be set or it will raise a 500 errors. At runtime the dependencies can check if the item as `default_assets` set so this will be pretty hard to enable `dynamic` check.
> 
> Note: this could be resolved using the Render extension at a collection level

Yes, I'm using this with titiler. I already needed custom dependencies for some other inputs so it was simple enough to add in the `default_assets`.

vincentsarago:
@mccarthyryanc thanks for the PR, I've made some edits:
- raise warnings for all the methods (it was implemented only for tile())
- add `default_bands` to the MultiBandReader

let's hope no one comes with the need of `default_expression` 😅

mccarthyryanc:
Awesome! Thanks for the changes @vincentsarago, those all look great!

Issue #717, opened 2024-08-14
Handle Xarray dataset without CRS


j08lue:
A pretty common case is that a DataArray passed to `XarrayReader` does not have a CRS defined (via `rioxarray`).

E.g. with the xarray tutorial data:

```python
#!pip install rio_tiler xarray rioxarray pooch netCDF4

import xarray
from rio_tiler.io.xarray import XarrayReader

ds = xarray.tutorial.open_dataset("air_temperature")
da = ds["air"]
print(da.rio.crs)  # -> None

with XarrayReader(da) as reader:
    img = reader.tile(0, 0, 0)
```

<details>

<summary>CRSError: CRS is invalid: None</summary>

```python
---------------------------------------------------------------------------
CRSError                                  Traceback (most recent call last)
File ~/Code/titiler-xarray/.venv/lib/python3.12/site-packages/rio_tiler/io/base.py:106, in SpatialMixin.tile_exists(self, tile_x, tile_y, tile_z)
    105 try:
--> 106     tile_bounds = transform_bounds(
    107         self.tms.rasterio_crs,
    108         self.crs,
    109         *tile_bounds,
    110         densify_pts=21,
    111     )
    112 except:  # noqa
    113     # HACK: gdal will first throw an error for invalid transformation
    114     # but if retried it will then pass.
    115     # Note: It might return `+/-inf` values

File ~/Code/titiler-xarray/.venv/lib/python3.12/site-packages/rasterio/warp.py:148, in transform_bounds(src_crs, dst_crs, left, bottom, right, top, densify_pts)
    147 src_crs = CRS.from_user_input(src_crs)
--> 148 dst_crs = CRS.from_user_input(dst_crs)
    149 return _transform_bounds(
    150     src_crs,
    151     dst_crs,
   (...)
    156     densify_pts,
    157 )

File rasterio/crs.pyx:783, in rasterio.crs.CRS.from_user_input()

CRSError: CRS is invalid: None

During handling of the above exception, another exception occurred:

CRSError                                  Traceback (most recent call last)
Cell In[8], line 11
      8 print(da.rio.crs)  # -> None
     10 with XarrayReader(da) as reader:
---> 11     img = reader.tile(0, 0, 0)

[...]

PR #725, opened 2024-09-02
[XARRAY] add env settings to fallback to WGS84 when no CRS is provided

vincentsarago:
closes https://github.com/cogeotiff/rio-tiler/issues/717

cc @j08lue

### DIFF
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import warnings
 from typing import Any, Dict, List, Optional
 
@@ -15,7 +16,12 @@ from rasterio.warp import calculate_default_transform
 from rasterio.warp import transform as transform_coords
 
 from rio_tiler.constants import WEB_MERCATOR_TMS, WGS84_CRS
-from rio_tiler.errors import PointOutsideBounds, TileOutsideBounds
+from rio_tiler.errors import (
+    MissingCRSWarning,
+    OverwritingCRSWarning,
+    PointOutsideBounds,
+    TileOutsideBounds,
+)
 from rio_tiler.io.base import BaseReader
 from rio_tiler.models import BandStatistics, ImageData, Info, PointData
 from rio_tiler.types import BBox, NoData, WarpResampling
@@ -30,6 +36,11 @@ try:
     import rioxarray
 except ImportError:  # pragma: nocover
     rioxarray = None  # type: ignore
+
+
+default_to_wgs84_crs = os.environ.get(
+    "RIO_TILER_XARRAY_DEFAULT_WGS84", "FALSE"
+).upper() in ["TRUE", "YES"]
 
 
 @attr.s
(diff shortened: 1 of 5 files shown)


rio_tiler/io/xarray.py
```diff
@@ -71,7 +82,20 @@ def __attrs_post_init__(self):
...
+                OverwritingCRSWarning,
+            )
+            self.input.rio.write_crs("epsg:4326", inplace=True)
+
         self.crs = self.input.rio.crs
+        if not self.crs:
+            warnings.warn(
+                "Dataset doesn't have a valid CRS set. rio-tiler might not work properly",
+                MissingCRSWarning,
+            )
```
j08lue, line 98: Why do we not throw an exception in this case?

Using an `XarrayTiler` for anything meaningful without a (valid) CRS, will inevitably fail.

rio_tiler/io/xarray.py
```diff
@@ -71,7 +82,20 @@ def __attrs_post_init__(self):
         assert rioxarray is not None, "rioxarray must be installed to use XarrayReader"
 
         self.bounds = tuple(self.input.rio.bounds())
+
+        if not self.input.rio.crs and default_to_wgs84_crs:
+            warnings.warn(
+                "Dataset doesn't have a valid CRS set. Automatically setting CRS to WGS84",
+                OverwritingCRSWarning,
+            )
+            self.input.rio.write_crs("epsg:4326", inplace=True)
```
j08lue, line 91: This mutates the `self.input` Xarray object (DataArray).

If we do not copy what the user passed in earlier, this behaviour could be unexpected for a user.

If there is a chance that the XarrayReader mutates the input, should we perhaps always have it create a copy up-front?

j08lue (commented):
This new config option makes the behaviour nicely explicit.

I would probably recommend to pretty much all users to turn this WGS84 default option on, since rectilinear lat/lon grids are so very common for Xarray datasets (climate data in NetCDF), so it is a pretty safe guess. One could add some heuristics around it, too, but that is not rio-tiler material.

Had a few questions regarding the implementation, please see below.

rio_tiler/io/xarray.py
vincentsarago: We usually don't fail for this because you can still access the dataset 🤷 

Let me think about it

rio_tiler/io/xarray.py
vincentsarago: 🤔 good point

vincentsarago:
> I would probably recommend to pretty much all users to turn this WGS84 default option on, since rectilinear lat/lon grids are so very common for Xarray datasets (climate data in NetCDF), so it is a pretty safe guess. One could add some heuristics around it, too, but that is not rio-tiler material

🤔 

Well I usually don't like magic, meaning if would prefer that people know what they are doing instead of trying to patch for every issues. IMO not having CRS set in their dataset is their issue, and shouldn't be rio-tiler one.

As think what we do in titiler-xarray is better https://github.com/developmentseed/titiler-xarray/blob/dev/titiler/xarray/reader.py#L170-L172

rio_tiler/io/xarray.py
vincentsarago: 😬 sadly rio-tiler gets the input directly from the user so we can't really avoid `editing` the origin data array

vincentsarago:
in #726 I've proposed another approach which raises errors instead of trying to fix the dataset and then expect the `bad` dataset to be handled at application level

Issue #718, opened 2024-08-14
XarrayReader considers tiles out of bounds when longitudes > 180?


j08lue:
I created an `rio_tiler.io.xarray.XarrayReader` for a dataset that has longitude range 200 to 330 degrees East (which is not uncommon in NetCDF datasets) and tried to get a global tile at zoom level 0.

That would fail, saying the tile was out of bounds.

Once I subtracted 360 to get the longitudes into the -180, 180 range, it worked.

Should we handle datasets with wrapping longitude axes better in `XarrayReader`? Or document the tight criteria that files need to fulfill to work?

Here is a minimal example: https://gist.github.com/j08lue/6ec9ecc2a853e499e4fca98d8aa5fb78

The gist:

![image](https://github.com/user-attachments/assets/0bba6cde-66f7-4410-811c-63f692036926)

```python
with XarrayReader(da) as reader:
    img = reader.tile(0, 0, 0)
```

fails with

```python
---------------------------------------------------------------------------
TileOutsideBounds                         Traceback (most recent call last)
Cell In[6], line 2
      1 with XarrayReader(da) as reader:
----> 2     img = reader.tile(0, 0, 0)

File ~/Code/titiler-xarray/.venv/lib/python3.12/site-packages/rio_tiler/io/xarray.py:235, in XarrayReader.tile(self, tile_x, tile_y, tile_z, tilesize, resampling_method, reproject_method, auto_expand, nodata)
    232     reproject_method = resampling_method
    234 if not self.tile_exists(tile_x, tile_y, tile_z):
--> 235     raise TileOutsideBounds(f"Tile {tile_z}/{tile_x}/{tile_y} is outside bounds")
    237 ds = self.input
    238 if nodata is not None:

TileOutsideBounds: Tile 0/0/0 is outside bounds
```

j08lue:
I see we have some code in TiTiler-Xarray that makes sure datasets conform to the format we expect, e.g.

```python
    if crs == "epsg:4326" and (da.x > 180).any():
        # Adjust the longitude coordinates to the -180 to 180 range
        da = da.assign_coords(x=(da.x + 180) % 360 - 180)

        # Sort the dataset by the updated longitude coordinates
        da = da.sortby(da.x)
```

https://github.com/developmentseed/titiler-xarray/blob/2be19eb9905d94cd6ac21f73947ad9802f9e92cb/titiler/xarray/reader.py#L174-L179

[...]

PR #726, opened 2024-09-02
raise error when xarray dataset have wrong geo information

vincentsarago:
another solution for #718 and #717 

This is a bit more `classic` in a way that this PR adds checks for missing CRS or invalid Geographic bounds and then will raise an error directly when opening the dataset.

ref https://github.com/cogeotiff/rio-tiler/pull/725#issuecomment-2324702461

### DIFF
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -89,7 +89,9 @@ class XarrayReader(BaseReader):
             or self.bounds[2] > 180
             or self.bounds[3] > 90
         ):
-            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}. Must be within (-180, -90, 180, 90).")
+            raise InvalidGeographicBounds(
+                f"Invalid geographic bounds: {self.bounds}. Must be within (-180, -90, 180, 90)."
+            )
 
         self._dims = [
             d
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -15,7 +15,12 @@ from rasterio.warp import calculate_default_transform
 from rasterio.warp import transform as transform_coords
 
 from rio_tiler.constants import WEB_MERCATOR_TMS, WGS84_CRS
-from rio_tiler.errors import PointOutsideBounds, TileOutsideBounds
+from rio_tiler.errors import (
+    InvalidGeographicBounds,
+    MissingCRS,
+    PointOutsideBounds,
+    TileOutsideBounds,
+)
 from rio_tiler.io.base import BaseReader
 from rio_tiler.models import BandStatistics, ImageData, Info, PointData
 from rio_tiler.types import BBox, NoData, WarpResampling
@@ -72,6 +77,18 @@ class XarrayReader(BaseReader):
 
         self.bounds = tuple(self.input.rio.bounds())
         self.crs = self.input.rio.crs
+        if not self.crs:
+            raise MissingCRS(
+                "Dataset doesn't have CRS information, please add it before using rio-tiler (e.g. `ds.rio.write_crs('epsg:4326', inplace=True)`)"
+            )
+
+        if self.crs == WGS84_CRS and (
+            self.bounds[0] < -180
+            or self.bounds[1] < -90
+            or self.bounds[2] > 180
+            or self.bounds[3] > 90
+        ):
+            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}")
 
         self._dims = [
             d
--- a/rio_tiler/io/xarray.py
+++ b/rio_tiler/io/xarray.py
@@ -89,7 +89,7 @@ class XarrayReader(BaseReader):
             or self.bounds[2] > 180
             or self.bounds[3] > 90
         ):
-            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}")
+            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}. Must be within (-180, -90, 180, 90).")
 
         self._dims = [
             d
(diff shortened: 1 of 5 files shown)


tests/test_io_xarray.py
```diff
@@ -321,3 +322,53 @@ def test_xarray_reader_resampling():
...
+
+def test_xarray_reader_no_crs():
+    """Should raise MissingCRS."""
+    arr = numpy.arange(0.0, 33 * 35).reshape(1, 33, 35)
+    data = xarray.DataArray(
+        arr,
+        dims=("time", "y", "x"),
+        coords={
+            "x": list(range(-170, 180, 10)),
+            "y": list(range(-80, 85, 5)),
```
j08lue, line 335: just a nitpick - why not

```suggestion
            "x": numpy.arange(-170, 180, 10),
            "y": numpy.arange(-80, 85, 5),
```

rio_tiler/io/xarray.py
```diff
@@ -72,6 +77,18 @@ def __attrs_post_init__(self):
...
+            raise MissingCRS(
+                "Dataset doesn't have CRS information, please add it before using rio-tiler (e.g. `ds.rio.write_crs('epsg:4326', inplace=True)`)"
+            )
+
+        if self.crs == WGS84_CRS and (
+            self.bounds[0] < -180
+            or self.bounds[1] < -90
+            or self.bounds[2] > 180
+            or self.bounds[3] > 90
+        ):
```
j08lue, line 90: Wondering whether this is robust enough.

Gridded datasets (CDF) are often north-up, but rasters (GeoTIFF etc) often south-up (negative cell/pixel width in y-direction).

I see `self.bounds = tuple(self.input.rio.bounds())` and `rio.bounds` [returns](https://corteva.github.io/rioxarray/html/rioxarray.html#rioxarray.rioxarray.XRasterBase.bounds) `left, bottom, right, top`, possibly taken from an affine transform.

In that case, we could easily encounter datasets in WGS84, but where the y bounds are flipped, which is ok for Rasterio to construct tiles from, I assume, but will fail this check here.

This would make the condition here robust to that case:

```suggestion
        if self.crs == WGS84_CRS and (
            self.bounds[0] < -180
            or min(self.bounds[1], self.bounds[3]) < -90
            or self.bounds[2] > 180
            or max(self.bounds[1], self.bounds[3]) > 90
        ):
```

j08lue (commented):
IMO, the early failure here is nicer than the the magic + potential head scratching of https://github.com/cogeotiff/rio-tiler/pull/725. 👌

rio_tiler/io/xarray.py
vincentsarago: thanks @j08lue 🙏

rio_tiler/io/xarray.py
```diff
@@ -72,6 +77,18 @@ def __attrs_post_init__(self):
 
         self.bounds = tuple(self.input.rio.bounds())
         self.crs = self.input.rio.crs
+        if not self.crs:
+            raise MissingCRS(
+                "Dataset doesn't have CRS information, please add it before using rio-tiler (e.g. `ds.rio.write_crs('epsg:4326', inplace=True)`)"
+            )
+
+        if self.crs == WGS84_CRS and (
+            self.bounds[0] < -180
```
vincentsarago, line 86: should then this be ?
```suggestion
            min(self.bounds[0], self.bounds[2]) < -180
```

rio_tiler/io/xarray.py
```diff
@@ -72,6 +77,18 @@ def __attrs_post_init__(self):
...
+            or min(self.bounds[1], self.bounds[3]) < -90
+            or self.bounds[2] > 180
```
vincentsarago, line 88: same 
```suggestion
            or max(self.bounds[0], self.bounds[2]) > 180
```

rio_tiler/io/xarray.py
vincentsarago: well in fact, rioxarray will `order` the bounds https://github.com/corteva/rioxarray/blob/25fbbce8612ba9a38786a67dc56d24c7d8f79db7/rioxarray/rioxarray.py#L229-L253

j08lue: Ah, they are mixing up "minx/maxx" (ordered projected values) and "left/right" (image orientation) lingo, I think. But the code is clear.

rio_tiler/io/xarray.py
```diff
@@ -70,8 +75,21 @@ def __attrs_post_init__(self):
...
+                "Dataset doesn't have CRS information, please add it before using rio-tiler (e.g. `ds.rio.write_crs('epsg:4326', inplace=True)`)"
+            )
+
+        if self.crs == WGS84_CRS and (
+            self.bounds[0] < -180
+            or self.bounds[1] < -90
+            or self.bounds[2] > 180
+            or self.bounds[3] > 90
+        ):
+            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}")
```
j08lue, line 92:
```suggestion
            raise InvalidGeographicBounds(f"Invalid geographic bounds: {self.bounds}. Must be within (-180, -90, 180, 90.")
```

j08lue (approved):
This gives users more helpful error messages when CRS is missing or their bounds are outside the expected range, closing

- https://github.com/cogeotiff/rio-tiler/issues/717
- https://github.com/cogeotiff/rio-tiler/issues/718

I think this is preferable to the more auto-magic solution proposed in https://github.com/cogeotiff/rio-tiler/pull/725.

NB: I added a comment below to let the users know which range is expected, since this is a GDAL peculiarity - longitude values outside that range are perfectly fine elsewhere. Up to you whether you want to accept my suggestion.

PR #745, opened 2024-10-07
refactor attribute and remove TMS from BaseReader init, and remove min/maxzoom from info response

### LINKED ISSUE #744: replace `geographic_bounds` with `bounds` and removes  `maxzoom, minzoom` from `info()` response
Those variable depends on the TMS and on the geographic CRS, which then needs to be passed and rightfully configured when opening the file 

https://github.com/cogeotiff/rio-tiler/blob/bb33ffb7c0307c5d4f5ebc9061eac200e64b4326/rio_tiler/models.py#L60-L73

ref: https://github.com/developmentseed/titiler/pull/995#issuecomment-2394357026

vincentsarago:
closes #744 

This PR does:
- refactors the `SpatialMixin` base class 
   - removes `tms` attribute
   - removes `geographic_crs` attribute
   - removes `minzoom` and `maxzoom` properties
   -  add `get_zooms(tms: TileMatrixSet)` method 
   - change `geographic_bounds` property to method `geographic_bounds(crs: CRS)` 
   - update `tile_exists` and add `tms` argument 

- changes `.tile()` method and add `tms` argument 

- refactors the `info()` method
    - return dataset bounds (instead of geographic bounds)
    - add `crs` metadata (epsg or wkt)
 
- removes `models.SpatialInfo` and `models.Bounds`

- removes `minzoom/maxzoom` attributes in MultiBaseReader and `MultiBandReader`

- starts a migration guide

```python
# before
import morecantile

tms = morecantile.tms.get("WebMercatorQuad")
with Reader("cog.tif", tms=tms, geographic_crs=tms.rasterio_geographic_crs) as src:
    _ = src.tile_exists(0, 0, 0)
    _ = src.tile(0, 0, 0)
    _ = src.minzoom
    _ = src.maxzoom
    _ = src.geographic_bounds

    info = src.info()
    assert info.bounds == src.geographic_bounds
    assert info.minzoom is not None
    assert info.maxzoom is not None

# now
import morecantile

tms = morecantile.tms.get("WebMercatorQuad")
with Reader("cog.tif") as src:
    _ = src.tile_exists(0, 0, 0, tms=tms)
    _ = src.tile(0, 0, 0, tms=tms)
    _, _ = src.get_zooms(tms)
    _ = src.geographic_bounds(tms.rasterio_geographic_crs)
    
    info = src.info()
    assert info.bounds == src.bounds == src.dataset.bounds
    assert info.crs == src.crs == src.dataset.crs
    assert not hasattr(info, "minzoom")
    assert not hasattr(info, "maxzoom")
```

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -4,7 +4,6 @@ import abc
 import contextlib
 import re
 import warnings
-from functools import cached_property
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 
 import attr
@@ -39,8 +38,6 @@ class SpatialMixin:
 
     """
 
-    tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
-
     bounds: BBox = attr.ib(init=False)
     crs: CRS = attr.ib(init=False)
 
@@ -48,12 +45,9 @@ class SpatialMixin:
     height: Optional[int] = attr.ib(default=None, init=False)
     width: Optional[int] = attr.ib(default=None, init=False)
 
-    geographic_crs: CRS = attr.ib(init=False, default=WGS84_CRS)
-
-    @cached_property
-    def geographic_bounds(self) -> BBox:
+    def geographic_bounds(self, crs: CRS = WGS84_CRS) -> BBox:
         """Return dataset bounds in geographic_crs."""
-        if self.crs == self.geographic_crs:
+        if self.crs == crs:
             if self.bounds[1] > self.bounds[3]:
                 warnings.warn(
                     "BoundingBox of the dataset is inverted (minLat > maxLat).",
@@ -71,7 +65,7 @@ class SpatialMixin:
         try:
             bounds = transform_bounds(
                 self.crs,
-                self.geographic_crs,
+                crs,
                 *self.bounds,
                 densify_pts=21,
             )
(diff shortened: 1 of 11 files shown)


rio_tiler/io/base.py
```diff
@@ -39,21 +38,16 @@ class SpatialMixin:
...
 
     transform: Optional[Affine] = attr.ib(default=None, init=False)
     height: Optional[int] = attr.ib(default=None, init=False)
     width: Optional[int] = attr.ib(default=None, init=False)
 
-    geographic_crs: CRS = attr.ib(init=False, default=WGS84_CRS)
-
-    @cached_property
-    def geographic_bounds(self) -> BBox:
+    def geographic_bounds(self, crs: CRS = WGS84_CRS) -> BBox:
```
AndrewAnnex, line 48: I think the default CRS here should not be WGS84 but the CRS returned from `self.crs.geodetic_crs` As a asset _propable should_ have a CRS, the geographic crs here should be appropriate to the CRS of the dataset itself.

rio_tiler/io/rasterio.py
```diff
@@ -176,10 +161,13 @@ def _get_descr(ix):
         else:
             nodata_type = "None"
 
+        crs_string = (
+            f"EPSG:{self.crs.to_epsg()}" if self.crs.to_epsg() else self.crs.to_wkt()
```
AndrewAnnex, line 165: hmm why not opt instead to use a urn/uri rather than attempt a epsg conversion here? those should work with the other authorities and be relatively compact. Failing through to wkt makes sense but also think wkt2_2019 should be preferred as it's more verbose and complete.

rio_tiler/io/base.py
vincentsarago: @AndrewAnnex the problem is that rasterio CRS class doesn't have `geodetic_crs` 😬

vincentsarago:
I feel I went too far with this PR (thanks for a good night of 🛏️) 

I've taken another approach over https://github.com/cogeotiff/rio-tiler/pull/746

rio_tiler/io/base.py
AndrewAnnex: ahh well that would be a problem. But I have an idea and will post in the new issue

PR #746, opened 2024-10-08
remove geographic_crs and update info model

vincentsarago:
second take at #744 

This PR does: 
- removes min/max zoom from `info()` response  **breaking change**
- adds CRS (uri or wkt) in `info()` response
- replaces `geographic_bounds` with dataset bounds in `info()` response **breaking change**
- removes `geographic_bounds` property from Reader BaseClass  **breaking change**
- adds `get_geographic_bounds(crs: CRS)` method in BaseClass 
- removes `geographic_crs` attribute in BaseClass  **breaking change**

This PR is less `breaking` than #745 and achieve the same result

### DIFF
--- a/rio_tiler/io/base.py
+++ b/rio_tiler/io/base.py
@@ -15,7 +15,7 @@ from rasterio.crs import CRS
 from rasterio.rio.overview import get_maximum_overview_level
 from rasterio.warp import calculate_default_transform, transform_bounds
 
-from rio_tiler.constants import WEB_MERCATOR_TMS, WGS84_CRS
+from rio_tiler.constants import WEB_MERCATOR_TMS
 from rio_tiler.errors import (
     AssetAsBandError,
     ExpressionMixingWarning,
@@ -27,7 +27,7 @@ from rio_tiler.errors import (
 from rio_tiler.models import BandStatistics, ImageData, Info, PointData
 from rio_tiler.tasks import multi_arrays, multi_points, multi_values
 from rio_tiler.types import AssetInfo, BBox, Indexes
-from rio_tiler.utils import cast_to_sequence, normalize_bounds
+from rio_tiler.utils import CRS_to_uri, cast_to_sequence, normalize_bounds
 
 
 @attr.s
@@ -48,12 +48,9 @@ class SpatialMixin:
     height: Optional[int] = attr.ib(default=None, init=False)
     width: Optional[int] = attr.ib(default=None, init=False)
 
-    geographic_crs: CRS = attr.ib(init=False, default=WGS84_CRS)
-
-    @cached_property
-    def geographic_bounds(self) -> BBox:
-        """Return dataset bounds in geographic_crs."""
-        if self.crs == self.geographic_crs:
+    def get_geographic_bounds(self, crs: CRS) -> BBox:
+        """Return Geographic Bounds for a Geographic CRS."""
+        if self.crs == crs:
             if self.bounds[1] > self.bounds[3]:
                 warnings.warn(
                     "BoundingBox of the dataset is inverted (minLat > maxLat).",
@@ -69,12 +66,7 @@ class SpatialMixin:
             return self.bounds
 
         try:
-            bounds = transform_bounds(
-                self.crs,
-                self.geographic_crs,
-                *self.bounds,
-                densify_pts=21,
-            )
+            bounds = transform_bounds(self.crs, crs, *self.bounds, densify_pts=21)
         except:  # noqa
             warnings.warn(
                 "Cannot determine bounds in geographic CRS, will default to (-180.0, -90.0, 180.0, 90.0).",
@@ -1085,9 +1077,8 @@ class MultiBandReader(SpatialMixin, metaclass=abc.ABCMeta):
         bands_metadata = multi_values(bands, _reader, **kwargs)
 
         meta = {
-            "bounds": self.geographic_bounds,
-            "minzoom": self.minzoom,
-            "maxzoom": self.maxzoom,
+            "bounds": self.bounds,
+            "crs": CRS_to_uri(self.crs) or self.crs.to_wkt(),
         }
 
         # We only keep the value for the first band.
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -789,3 +789,21 @@ def cast_to_sequence(val: Optional[Any] = None) -> Sequence:
         val = (val,)
 
     return val
+
+
+def CRS_to_uri(crs: CRS) -> Optional[str]:
+    """Convert CRS to URI.
+
+    Code adapted from https://github.com/developmentseed/morecantile/blob/1829fe12408e4a1feee7493308f3f02257ef4caf/morecantile/models.py#L148-L161
+    """
+    # attempt to grab the authority, version, and code from the CRS
+    if authority_code := crs.to_authority(confidence_threshold=70):
+        version = "0"
+        authority, code = authority_code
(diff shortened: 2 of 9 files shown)


vincentsarago:
with this PR, a `tilejson.json` endpoint in titiler will look like

```python

@app.get(
    "/{tileMatrixSetId}/tilejson.json",
    response_model=TileJSON,
    responses={200: {"description": "Return a tilejson"}},
    response_model_exclude_none=True,
)
def tilejson(
    request: Request,
    tileMatrixSetId: Annotated[
        Literal[tuple(self.supported_tms.list())],
        Path(
            description="Identifier selecting one of the TileMatrixSetId supported."
        ),
    ],
):
    """Return TileJSON document for a dataset."""
    route_params = {
        "z": "{z}",
        "x": "{x}",
        "y": "{y}",
        "tileMatrixSetId": tileMatrixSetId,
    }
    tiles_url = self.url_for(request, "tile", **route_params)

    tms = self.supported_tms.get(tileMatrixSetId)
    
    with Reader("COG.tif", tms=tms) as src_dst:
        return {
            "bounds": src_dst.get_geographic_bounds(tms.rasterio_geographic_crs),
            "minzoom": src_dst.minzoom,
            "maxzoom": src_dst.maxzoom,
            "tiles": [tiles_url],
        }
```

rio_tiler/io/base.py
```diff
@@ -48,12 +48,9 @@ class SpatialMixin:
     height: Optional[int] = attr.ib(default=None, init=False)
     width: Optional[int] = attr.ib(default=None, init=False)
 
-    geographic_crs: CRS = attr.ib(init=False, default=WGS84_CRS)
-
-    @cached_property
-    def geographic_bounds(self) -> BBox:
-        """Return dataset bounds in geographic_crs."""
-        if self.crs == self.geographic_crs:
+    def get_geographic_bounds(self, crs: CRS) -> BBox:
```
vincentsarago, line 51: `get_geographic_bounds` or `geographic_bounds`?

vincentsarago, line 51: should we default to `EPSG:4326`?

AndrewAnnex: on balance `get_geographic_bounds` as it's no longer a cached property, but it's just a preference and not a huge concern to me

vincentsarago:
would love to have your opinions @AndrewAnnex 🙏

AndrewAnnex:
@vincentsarago heh the time zones probably make this hard but just added a suggestion

rio_tiler/utils.py
```diff
@@ -789,3 +789,21 @@ def cast_to_sequence(val: Optional[Any] = None) -> Sequence:
         val = (val,)
 
     return val
+
+
+def CRS_to_uri(crs: CRS) -> Optional[str]:
+    """Convert CRS to URI.
+
+    Code adapted from https://github.com/developmentseed/morecantile/blob/1829fe12408e4a1feee7493308f3f02257ef4caf/morecantile/models.py#L148-L161
```
AndrewAnnex, line 797: isn't morecantile a dependency of rio-tiler?

rio_tiler/io/base.py
AndrewAnnex: oh but wait the tms is defined here, sorry I think I am confusing myself a bit.
let me think a bit more

AndrewAnnex: okay I am a bit confused here, the spatial mixin provides the `tms` as a property so why not use the TMS's rasterio geographic_crs here? If the `crs` should be used rather than the TMS then what I said above would be the right call, as I could imagine having a cog or some other asset with a particular CRS and then use a TMS with a different projection (and potentially different geographic_crss in weird edge cases for planets). Seems like either way is a easy path, but maybe I am missing something @vincentsarago ?

vincentsarago: the `tms` and the `geographic_bounds` are not directly linked together. In a sense that a user could want the geographic bounds of a dataset without knowing or caring what TMS is used.

I've written an exampling showing how the geographic crs from the TMS could be used in a tilejson endpoint  https://github.com/cogeotiff/rio-tiler/pull/746#issuecomment-2399279706

vincentsarago: I would love to avoid having both pyproj and rasterio CRS code here. we've done in in morecantile to accommodate rio-tiler but I don't think we should use it more than what's used right now.

Note: pyproj and rasterio comes with wheel which might have different PROJ version so mixing them is not 💯 safe

rio_tiler/utils.py
AndrewAnnex: sure? I am just saying you don't have to copy the code over since it's already a dependency so you can just import it. I don't think there is ever a situation where morecantile would not be installed with rio-tiler

vincentsarago: This version is also a bit different

rio_tiler/io/base.py
AndrewAnnex: @vincentsarago I am not convinced that mixing rasterio with pyproj is that problematic, but then again I tend to use conda-forge distributions of these and so mixing gdal/pyproj versions isn't something I've had issues with before. 

Okay it then sounds like the `.crs` must be used. To do so  we'd need implement some logic either within rio_tiler to construct the geographic crs from the `.data` attribute of the rasterio crs if it's projected, or we'd need to update rasterio's CRS class to use something like the `CloneGeogCS` function to extract it. I am going to look into that now as it's a good idea for rasterio to have this

AndrewAnnex: but that said I don't think that needs to block this PR, so long as we leave it as an explicit parameter that must be provided until some good solution is found for the above without using pyproj

vincentsarago: @AndrewAnnex I'm not quite sure to understand what's your usage of the `geographic_bounds` property. 

If it's for titiler, I think the example I've shared, shows how to get the correct geographic bounds by using the TMS geographic CRS. 

If it is for some other thing, you can easily create a custom function that takes a rasterio CRS (from `Reader().crs`) and convert it to a pyproj CRS to get the `geodetic_crs`. 

As far as rio-tiler is concerned, removing the `geographic_crs` attribute and the new `get_geographic_bounds` method should resolve most of the pain points mentioned in the different issues.

AndrewAnnex: @vincentsarago the intent is so that already existing, large collections of analysis ready data provided by the USGS for the Moon and Mars can be used (see https://stac.astrogeology.usgs.gov/docs/). 

For the moment though yes I think the usage here is primarily concerned with titiler and the rest of the titiler stack of projects (morecantile, rio-cogeo, cogeo-mosaic, tilebench, etc). So any places where WGS84 gets assumed are problematic hence the other prs I've been making. 

But I've been a bit tunnel-visioned with this vendetta against hard coded CRSs, so I think you are correct that the above method does the job, and changing the api this way may cause other changes to be necessary, but I don't have a complete picture of it all in my head at the moment. 

My comments were made with an eye to minimizing PRs needed elsewhere, but maybe I am just over thinking it.

vincentsarago: > For the moment though yes I think the usage here is primarily concerned with titiler and the rest of the titiler/morecanitle stack of projects (morecantile, rio-cogeo, cogeo-mosaic, tilebench, etc). So any places where WGS84 gets assumed are problematic hence the other prs I've been making.

The next version of rio-tiler is going to be a major version, with other breaking change so all the projects will be updated to match the new version

titiler has also a lot of breaking changes! 

I do believe that removing the geographic_crs attribute is a good thing!

PR #752, opened 2024-10-22
Add CRS_to_urn

AndrewAnnex:
new `CRS_to_urn` function that uses a helper method made from old `CRS_to_uri` function (which is still provided)

did this in github.dev so didn't format with black (oops)

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -791,7 +791,7 @@ def cast_to_sequence(val: Optional[Any] = None) -> Sequence:
     return val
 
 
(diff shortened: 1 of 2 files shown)


rio_tiler/utils.py
```diff
@@ -791,7 +791,7 @@ def cast_to_sequence(val: Optional[Any] = None) -> Sequence:
     return val
 
 
-def CRS_to_uri(crs: CRS) -> Optional[str]:
+def _CRS_authority_info(crs: CRS) -> Optional[Tuple[str, str, str]]:
```
vincentsarago, line 794: Let's make this method private so we can break it when ever we want 🙈

Issue #763, opened 2024-11-08
Cast data to colormap datatype


vincentsarago:
We could cast the input data array to the datatype of the colormap to make sure we don't get issues like https://github.com/developmentseed/titiler/issues/1023

https://github.com/cogeotiff/rio-tiler/blob/dd5ce03f8f463ec93f3a8787d7155d318397cd7c/rio_tiler/colormap.py#L122-L130

PR #766, opened 2024-11-14
cast data to uint8

vincentsarago:
closes https://github.com/cogeotiff/rio-tiler/issues/763


rio_tiler/colormap.py
```diff
@@ -125,17 +126,19 @@ def apply_cmap(data: numpy.ndarray, colormap: ColorMapType) -> DataMaskType:
...
+    lookup_table = make_lut(colormap)  # type: ignore
     data = lookup_table[data[0], :]
 
     data = numpy.transpose(data, [2, 0, 1])
 
-    # If the colormap has values between 0-255
-    # we cast the output array to Uint8.
-    if data.min() >= 0 and data.max() <= 255:
-        data = data.astype("uint8")
-
```
vincentsarago, line 138: The data will be Uint8 and the Lut is also uint8 so there is no need for this

Issue #774, opened 2025-01-02
valid_percent in utils/get_array_statistics not correctly calculated, when using feature coverage array


MarcelCode:
If I calculate the coverage based on a geometry and hand it over to the get_array_statistics method, I would expect that the valid_percent is calculated based on the coverage array instead of the bounding box of the geometry.

![example_geometry](https://github.com/user-attachments/assets/cc3ab822-1bb5-4511-85bc-9c076df38591)

In this example valid_pixels is correctly calculated, but valid_percent is wrong: 
- 2 valid pixels, because one pixel is nodata. 
- As the geometry includes three pixels, I would expect that valid_percent **needs to be 66.67 %, instead it is 50 %.**  

Here is a full code running example:
```python
from rio_tiler.io import Reader

def calculate_statistics(filename: str, shape: dict) -> dict:
    with Reader(filename) as src:
        data = src.feature(shape)

        coverage_array = data.get_coverage_array(shape)

        return data.statistics(coverage=coverage_array)

if __name__ == "__main__":
    file = "https://rastless-tests.s3.eu-central-1.amazonaws.com/TUR_us-newyork_013032_EOMAP_20190424_153304_LSAT8_m0030_32bit.tif"

    aoi = {"type": "Polygon", "coordinates": [
        [[-74.046461218997848, 40.63946290226923], [-74.046464353686744, 40.638997869937697],
         [-74.045818607775814, 40.638991923211712], [-74.045817040431388, 40.639200058305974],
         [-74.046160288864613, 40.639204815671974], [-74.046161856209054, 40.63946290226923],
         [-74.046461218997848, 40.63946290226923]]]}


    statistics = calculate_statistics(file, aoi)

    valid_pixels = statistics["b1"]["valid_pixels"]
    valid_percent = statistics["b1"]["valid_percent"]

    print(f"Valid pixels: {valid_pixels}. Expected value: 2.0")
    print(f"Valid percent: {valid_percent} %. Expected value: 66.67 %")
```

From what I see in the code this line seems to be wrong, as it always calculates the valid_percent based on the full array size:
utils.py/get_array_statistics line 136
```python
valid_percent = round((valid_pixels / data[b].size) * 100, 2)
```

My suggestion to change it to:
```python
valid_percent = round((valid_pixels / np.count_nonzero(coverage)) * 100, 2)
```

Even if no coverage is handed over by the user the calculation is correct, because then it is created inside the method. 

I would be happy to create a PR if you agree.

[...]

PR #775, opened 2025-01-05
Fix issue #774: Use coverage array for calculation of valid_percent

MarcelCode:
**Summary**
This PR updates the calculation of valid_percent in get_array_statistics. Previously, valid_percent was calculated by dividing the number of valid pixels by the total size of the array. Now, it uses the sum of nonzero values in the coverage array for more accurate results.

**Why this change?**
The previous method incorrectly calculated valid_percent when the coverage array was created using a geometry that wasn’t a simple rectangle. This change ensures the calculation properly reflects the valid area within the intended geometry.

**Related Issue**
See issue #774 for a detailed explanation of the problem and the motivation for this change.

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -133,6 +133,8 @@ def get_array_statistics(
     else:
         coverage = numpy.ones((data.shape[1], data.shape[2]))
 
+    coverage_pixels = numpy.count_nonzero(coverage)
+
     # Avoid non masked nan/inf values
     numpy.ma.fix_invalid(data, copy=False)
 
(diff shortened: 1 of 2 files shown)


vincentsarago:
@MarcelCode thanks for the PR 🙏 

Because it's based on the `master` branch which has some `breaking change` (dropping python 3.8 support), If I wanted to release this change `as is`, I would need to publish a `7.3.0` version (minor version update), would this be fine for you? 

I can backport this fix in a `7.2.3` if needed

MarcelCode:
I think a new minor version is fine. No need to backport. 
I will pull the new version for our current projects, when it is done.

vincentsarago:
Thanks @MarcelCode 
I'll wait a few days to see if we can get https://github.com/cogeotiff/rio-tiler/pull/764 merged before doing a new release 

hope that's fine

PR #782, opened 2025-01-25
Reproject method for ImageData

emmanuelmathot:
**Proposed Changes:**

This PR add reproject helper to ImageData. a method that is necessary for https://github.com/sentinel-hub/titiler-openeo/issues/15 but I believe interesting to have directly in rio-tiler

**PR Checklist:**

- [x] Tests pass 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "morecantile>=5.0,<7.0",
     "pydantic~=2.0",
     "pystac>=0.5.4",
-    "rasterio>=1.4.2",
+    "rasterio>=1.4.0",
     "color-operations",
     "typing-extensions",
     "importlib_resources>=1.1.0; python_version < '3.9'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "morecantile>=5.0,<7.0",
     "pydantic~=2.0",
     "pystac>=0.5.4",
-    "rasterio>=1.3.0",
+    "rasterio>=1.4.2",
     "color-operations",
     "typing-extensions",
     "importlib_resources>=1.1.0; python_version < '3.9'",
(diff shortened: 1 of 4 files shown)


vincentsarago:
FYI the CI fails but it's because of a bug in pystac (probably fixed by https://github.com/stac-utils/pystac/pull/1505)

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "morecantile>=5.0,<7.0",
     "pydantic~=2.0",
     "pystac>=0.5.4",
-    "rasterio>=1.3.0",
+    "rasterio>=1.4.0",
```
vincentsarago, line 32: https://github.com/rasterio/rasterio/blob/main/CHANGES.txt#L107-L109

rio_tiler/models.py
```diff
@@ -786,3 +787,44 @@ def get_coverage_array(
         ).astype("float32")
 
         return cover_array.sum(-1).sum(1) / (cover_scale**2)
+
+    def reproject(
+        self,
+        dst_crs: CRS,
+        resolution: Optional[Tuple[float, float]] = None,
+        reproject_method: WarpResampling = "nearest",
```
vincentsarago, line 795: changed this to match what we use in the other methods

rio_tiler/models.py
```diff
@@ -786,3 +787,44 @@ def get_coverage_array(
...
+            dst_crs,
+            self.width,
+            self.height,
+            *self.bounds,
+            resolution=resolution,
+        )
+
+        destination = numpy.ma.masked_array(
+            numpy.zeros((self.count, h, w), dtype=self.array.dtype),
+        )
```
vincentsarago, line 809: we don't need to initialize the mask itself

rio_tiler/models.py
```diff
@@ -786,3 +787,44 @@ def get_coverage_array(
...
+            self.array,
+            destination,
+            src_transform=self.transform,
+            src_crs=self.crs,
+            dst_transform=dst_transform,
+            dst_crs=dst_crs,
+            resampling=Resampling[reproject_method],
+        )
+
+        bounds = array_bounds(h, w, dst_transform)
```
vincentsarago, line 820: change: avoid doing re-projection but use the transform to calculate the bounds

vincentsarago (approved):

PR #794, opened 2025-03-19
add interpolate=True/False to point methods

### LINKED ISSUE #793: Reader's `Point` method vs GDAL `gdallocationinfo`
Ref: https://github.com/developmentseed/titiler/discussions/1104

We have major difference in our `Reader.point(Lon, Lat)` method with GDAL `gdallocationinfo` method, as explained in https://github.com/developmentseed/titiler/discussions/1104#discussioncomment-12534469

GDAL take the partial coordinates and read a 0.5 pixel buffer around it while we take the `rounded` (with `numpy.floor`) pixel coordinates to get the value 

```python
import numpy
import rasterio
from rasterio import windows

row, col = 223.1128, 240.8924 # given in https://github.com/developmentseed/titiler/discussions/1104#discussion-8092649

with rasterio.open("https://cdn.proj.org/us_nga_egm96_15.tif") as src:
    # GDAL 
    window = windows.Window(row_off=row - 0.5, col_off=col - 0.5, width=1, height=1)
    print(src.read(window=window))

    # rio-tiler
    row = numpy.floor(row)
    col = numpy.floor(col)
    window = windows.Window(row_off=row, col_off=col, width=1, height=1)
    print(src.read(window=window))
```

Image

vincentsarago:
closes #793 

```python

lon, lat = -119.9019, 34.3468
with Reader("https://cdn.proj.org/us_nga_egm96_15.tif") as src:
    print(src.point(lon, lat, coord_crs="epsg:4326").data)
    print(src.point(lon, lat, coord_crs="epsg:4326", resampling_method="bilinear", interpolate=True).data)

[-36.604877]
[-35.990685]
```

cc @scottyhq

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -11,7 +11,7 @@ from rasterio import windows
 from rasterio.crs import CRS
 from rasterio.enums import ColorInterp, Resampling
 from rasterio.io import DatasetReader, DatasetWriter
-from rasterio.transform import array_bounds
+from rasterio.transform import array_bounds, rowcol
 from rasterio.vrt import WarpedVRT
 from rasterio.warp import transform as transform_coords
 from rasterio.warp import transform_bounds
@@ -508,6 +508,7 @@ def point(
     vrt_options: Optional[Dict] = None,
     resampling_method: RIOResampling = "nearest",
     reproject_method: WarpResampling = "nearest",
+    interpolate: bool = False,
     unscale: bool = False,
     post_process: Optional[Callable[[numpy.ma.MaskedArray], numpy.ma.MaskedArray]] = None,
 ) -> PointData:
@@ -522,6 +523,7 @@ def point(
         vrt_options (dict, optional): Options to be passed to the rasterio.warp.WarpedVRT class.
         resampling_method (RIOResampling, optional): RasterIO resampling algorithm. Defaults to `nearest`.
         reproject_method (WarpResampling, optional): WarpKernel resampling algorithm. Defaults to `nearest`.
+        interpolate (bool, optional): Interpolate pixels around the coordinates. Defaults to `False`.
         unscale (bool, optional): Apply 'scales' and 'offsets' on output data value. Defaults to `False`.
         post_process (callable, optional): Function to apply on output data and mask values.
 
@@ -580,7 +582,15 @@ def point(
         ):
             raise PointOutsideBounds("Point is outside dataset bounds")
 
-        row, col = dataset.index(lon, lat)
+        if interpolate:
+            # Ref: https://github.com/cogeotiff/rio-tiler/issues/793
+            # https://github.com/OSGeo/gdal/blob/a3d68b069e6b3676ba437faca5dca6ae2076ce24/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py#L185-L197
+            rows, cols = rowcol(dataset.transform, xs=[lon], ys=[lat], op=lambda x: x)
+            row = rows[0] - 0.5
+            col = cols[0] - 0.5
+        else:
+            row, col = dataset.index(lon, lat)
+
         img = read(
             dataset,
             indexes=indexes,
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -423,6 +423,33 @@ def test_point():
         assert pt.mask == numpy.array([0])
         assert pt.band_names == ["b1"]
 
+        # Interpolate=True but with Nearest, so no influence
+        pt = reader.point(
+            src_dst,
+            [-57.566, 73.6885],
+            coord_crs="epsg:4326",
+            indexes=1,
+            nodata=1,
+            interpolate=True,
+        )
+        assert pt.data == numpy.array([2800])
+        assert pt.mask == numpy.array([255])
+        assert pt.band_names == ["b1"]
+
+        # Interpolate=True + resampling=Cubic
+        pt = reader.point(
+            src_dst,
+            [-57.566, 73.6885],
+            coord_crs="epsg:4326",
+            indexes=1,
+            nodata=1,
+            resampling_method="cubic",
+            interpolate=True,
+        )
+        assert pt.data == numpy.array([2812])
+        assert pt.mask == numpy.array([255])
+        assert pt.band_names == ["b1"]
+
     with rasterio.open(COG_SCALE) as src_dst:
         pt = reader.point(src_dst, [310000, 4100000], coord_crs=src_dst.crs, indexes=1)
         assert pt.data == numpy.array([8917])
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 # Unreleased
 
 * use `GDAL_MEM_ENABLE_OPEN=TRUE` when opening a numpy array with rasterio
+* add `interpolate=True/False` to `.point()` methods to allow interpolation of surrounding pixels
 
 # 7.5.0 (2025-02-26)
 


rio_tiler/reader.py
```diff
@@ -580,7 +582,15 @@ def point(
         ):
             raise PointOutsideBounds("Point is outside dataset bounds")
 
-        row, col = dataset.index(lon, lat)
+        if interpolate:
+            # Ref: https://github.com/cogeotiff/rio-tiler/issues/793
+            # https://github.com/OSGeo/gdal/blob/a3d68b069e6b3676ba437faca5dca6ae2076ce24/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py#L185-L197
+            rows, cols = rowcol(dataset.transform, xs=[lon], ys=[lat], op=lambda x: x)
```
vincentsarago, line 588: we use `op=lambda x: x` to make sure rasterio doesn't apply any `rounding` on the pixel coordinates

FYI we can't use `dataset.index(lon, lat, op=lambda x: x)` because rasterio will use `int()` https://github.com/rasterio/rasterio/blob/main/rasterio/transform.py#L135 😓

vincentsarago:
Q: do we really need `interpolate`? 

maybe `nearest` will always give the correct result 🤷

vincentsarago:
answering my own question: NO

```python
import rasterio
from rasterio.warp import transform as transform_coords
from rasterio import windows

from rasterio.enums import Resampling

row, col = 223, 240 # given in https://github.com/developmentseed/titiler/discussions/1104#discussion-8092649

with rasterio.open("https://cdn.proj.org/us_nga_egm96_15.tif") as src:
    for off in range(0, 10):
        print("offset: ", off/10., "value: ",
            src.read(
                window=windows.Window(row_off=row - off/10., col_off=col - off/10., width=1, height=1),
                masked=True,
                resampling=Resampling["nearest"],
            )[0][0]
        )

offset:  0.0 value:  [-36.604877]
offset:  0.1 value:  [-36.604877]
offset:  0.2 value:  [-36.604877]
offset:  0.3 value:  [-36.604877]
offset:  0.4 value:  [-36.604877]
offset:  0.5 value:  [-36.604877]
offset:  0.6 value:  [-36.141163]
offset:  0.7 value:  [-36.141163]
offset:  0.8 value:  [-36.141163]
offset:  0.9 value:  [-36.141163]
```

Screenshot 2025-03-19 at 1 23 03 PM

vincentsarago:
@scottyhq 
🤦  I was doing `row_off=row - off/10., col_off=col - off/10.` instead of `row_off=row + off/10., col_off=col + off/10.`

```
import rasterio
from rasterio.warp import transform as transform_coords
from rasterio import windows

from rasterio.enums import Resampling

row, col = 223, 240 # given in https://github.com/developmentseed/titiler/discussions/1104#discussion-8092649

with rasterio.open("https://cdn.proj.org/us_nga_egm96_15.tif") as src:
    for off in range(0, 10):
        print("offset: ", off/10., "value: ",
            src.read(
                window=windows.Window(row_off=row + off/10., col_off=col + off/10., width=1, height=1),
                masked=True,
                resampling=Resampling["nearest"],
            )[0][0]
        )

offset:  0.0 value:  [-36.604877]
offset:  0.1 value:  [-36.604877]
offset:  0.2 value:  [-36.604877]
offset:  0.3 value:  [-36.604877]
offset:  0.4 value:  [-36.604877]
offset:  0.5 value:  [-37.332676]
offset:  0.6 value:  [-37.332676]
offset:  0.7 value:  [-37.332676]
offset:  0.8 value:  [-37.332676]
offset:  0.9 value:  [-37.332676]
```

scottyhq:
> 🤦 I was doing row_off=row - off/10. 

Ah, well, I didn't notice that either :) 

> read a 1pixel buffer around the xy coordinates (using the provided resampling
> interpolate=True: match gdallocationinfo and enable bilinear resampling 

I see. Are you only wanting to support nearest/bilinear/average then? Or perhaps also introduce a `buffer=1` arg, such that you could customize 'average' sampling, or if 'cubic' is requests set `buffer=4`?

vincentsarago:
@scottyhq 

>  Are you only wanting to support nearest/bilinear/average then?

it should support all resampling. I'll add a buffer option 👍

vincentsarago:
we should be good to merge this now @scottyhq

rio_tiler/reader.py
```diff
@@ -531,6 +535,11 @@ def point(
     """
     indexes = cast_to_sequence(indexes)
 
+    if buffer and buffer % 0.5:
+        raise InvalidBufferSize(
+            "`buffer` must be a multiple of `0.5` (e.g: 0.5, 1, 1.5, ...)."
+        )
```
vincentsarago, line 541: we use the same definition for buffer (floating point, multiple of 0.5) from the other methods

rio_tiler/reader.py
```diff
@@ -580,11 +589,37 @@ def point(
...
+            if buffer:
+                warnings.warn(
+                    "'buffer' will be ignored when `interpolate=False`.", UserWarning
+                )
+
+            if resampling_method != "nearest":
+                warnings.warn(
+                    f"{resampling_method} resampling will be ignored when `interpolate=False`.",
+                    UserWarning,
+                )
```
vincentsarago, line 614: add some warning if resampling/buffer is used without interpolate=True

tests/test_reader.py
```diff
@@ -423,6 +423,114 @@ def test_point():
...
+            src_dst,
+            [-57.566, 73.6885],
+            coord_crs="epsg:4326",
+            indexes=1,
+            nodata=1,
+            resampling_method="cubic",
+            interpolate=True,
+            buffer=0.5,
+        )
+        assert pt.data == numpy.array([2812])
```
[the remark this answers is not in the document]
vincentsarago: 🤷 I'm not sure, GDAL/Rasterio do not complain about this 😅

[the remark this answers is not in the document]
scottyhq: There must be some magic happening, I'll try to do some more digging!
https://github.com/OSGeo/gdal/blob/6c240a1451c5aba06fef698be5136c3e4caa393b/alg/gdalwarper.h#L37-L42

rio_tiler/reader.py
```diff
@@ -580,11 +589,37 @@ def point(
         ):
             raise PointOutsideBounds("Point is outside dataset bounds")
 
-        row, col = dataset.index(lon, lat)
+        if interpolate:
+            # Ref: https://github.com/cogeotiff/rio-tiler/issues/793
+            # https://github.com/OSGeo/gdal/blob/a3d68b069e6b3676ba437faca5dca6ae2076ce24/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py#L185-L197
```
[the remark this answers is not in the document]
vincentsarago: > thus the hardcoded sampling of only 4 pixels

where do you see this?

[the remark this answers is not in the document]
scottyhq: I guess 'up to 4 pixels' since it's just a window size of 1 (Like the schematic you show above)
https://github.com/OSGeo/gdal/blob/876952031b5118729206271e1108ab270ed0c067/swig/python/gdal-utils/osgeo_utils/samples/gdallocationinfo.py#L188-L191

tests/test_reader.py
```diff
@@ -423,6 +423,114 @@ def test_point():
...
+            src_dst,
+            [-57.566, 73.6885],
+            coord_crs="epsg:4326",
+            indexes=1,
+            nodata=1,
+            interpolate=True,
+        )
+        assert pt.data == numpy.array([2800])
+        assert pt.mask == numpy.array([255])
+        assert pt.band_names == ["b1"]
```
scottyhq, line 459: Does pt also keep track of array coordinates? e.g. `assert pt.xy = np.array([1090,1086])`
```
gdallocationinfo -r nearest tests/fixtures/cog.tif -wgs84 -57.566 73.68856
Report:
  Location: (1090P,1086L)
  Band 1:
    Value: 2800
```

tests/test_reader.py
```diff
@@ -423,6 +423,114 @@ def test_point():
...
+        pt = reader.point(
+            src_dst,
+            [-57.566, 73.6885],
+            coord_crs="epsg:4326",
+            indexes=1,
+            nodata=1,
+            resampling_method="bilinear",
+            interpolate=True,
+        )
+        assert pt.data == numpy.array([2819])
```
scottyhq, line 471: I think it makes sense to promote dtype from int->float for bilinear resampling:

```
gdallocationinfo -r bilinear  tests/fixtures/cog.tif -wgs84 -57.566 73.68856
Report:
  Location: (1090.59247446413P,1086.25414298276L)
  Band 1:
    Value: 2819.23468853823
```

tests/test_reader.py
vincentsarago: no but we could 👍

tests/test_reader.py
vincentsarago: 🤔 we currently don't expose `datatype` (`out_dtype`) 

I usually don't like when the API decide stuff for the user. We could add `out_dtype` option but I think it should be up to the user to ask for a specific output data type 

what do you think @scottyhq ?

scottyhq: Yeah there are pros and cons for sure. I do think there should be an option though - consider asking for the  midpoint between values 0 & 1, I would want/expect 0.5 as a result.

vincentsarago: @scottyhq #796 will take care of this 

are we good to go with this PR?

scottyhq: Sounds good. Though I think it'll be important to check that the resampling happens on promoted dtype, rather than interpolate on input dtype and finally cast to out_dtype.

vincentsarago:
@scottyhq 

🤦 I realize that the `buffer` option is wrong 

we can't use buffer as implemented right now because it will return a window of 4x4 values (if buffer=2) while `point` should only return one value https://github.com/cogeotiff/rio-tiler/blob/a28d944edbfd668c04e7feefd1f81b5d64991978/rio_tiler/reader.py#L630

I'll remove the `buffer` option.

vincentsarago:
```python

lon, lat = -119.875, 34.125
with Reader("https://cdn.proj.org/us_nga_egm96_15.tif") as src:
    print(src.point(lon, lat, coord_crs="epsg:4326").data)
    >>> [-37.332676]
    
    print(src.point(lon, lat, coord_crs="epsg:4326", resampling_method="bilinear", interpolate=True).data)
    >>> [-36.87111]

    print(src.point(lon, lat, coord_crs="epsg:4326", resampling_method="cubic", interpolate=True).data)
    >>> [-36.945553]

    print(src.point(lon, lat, coord_crs="epsg:4326", resampling_method="cubic_spline", interpolate=True).data)
    >>> [-36.8468]


gdallocationinfo -r nearest /vsicurl/https://cdn.proj.org/us_nga_egm96_15.tif -geoloc -119.875 34.125
-37.3326759338379

gdallocationinfo -r bilinear /vsicurl/https://cdn.proj.org/us_nga_egm96_15.tif -geoloc -119.875 34.125
-36.8711071014404

gdallocationinfo -r cubic /vsicurl/https://cdn.proj.org/us_nga_egm96_15.tif -geoloc -119.875 34.125
-36.9455545693636

gdallocationinfo -r cubicspline /vsicurl/https://cdn.proj.org/us_nga_egm96_15.tif -geoloc -119.875 34.125
-36.8468017925819
```

We know get the same result as `gdallocationinfo` so I'll merge (after adding the `pixel_location` info in the output model)

tests/test_reader.py
vincentsarago: We'll check this when implementing the `out_dtype` option 🙏

PR #803, opened 2025-05-22
force GCPS dataset to be in a second WarpedVRT dataset to allow boundless read

vincentsarago:
closes https://github.com/developmentseed/titiler/issues/1163

### DIFF
--- a/rio_tiler/reader.py
+++ b/rio_tiler/reader.py
@@ -308,7 +308,7 @@ def read(
             data,
             bounds=out_bounds,
             crs=dataset.crs,
-            band_names=[f"b{idx}" for idx in indexes],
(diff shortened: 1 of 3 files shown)


rio_tiler/reader.py
```diff
@@ -308,7 +308,7 @@ def read(
             data,
             bounds=out_bounds,
             crs=dataset.crs,
-            band_names=[f"b{idx}" for idx in indexes],
+            band_names=[dataset.descriptions[ix - 1] or f"b{idx}" for idx in indexes],
```
vincentsarago, line 311: oh nooooo that shouldn't be there

I need to yank 7.7.3 release

PR #813, opened 2025-06-24
add include/exclude drivers for rasterio Reader

vincentsarago:


vincentsarago:
better use `GDAL_SKIP` https://gdal.org/en/stable/user/configoptions.html#driver-management

PR #875, opened 2026-03-04
remove useless async methods

vincentsarago:
This PR removes `async` context manager (replace with sync context manager for compatibility with Sync Reader) and also remove `async` definition for `info()` methods which doesn't require it 

This will help the integration in titiler, and reduce the number custom code


rio_tiler/experimental/_async.py
```diff
@@ -53,16 +53,16 @@ class AsyncBaseReader(SpatialMixin, metaclass=abc.ABCMeta):
     input: Any = attr.ib()
     tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
 
-    async def __aenter__(self):
```
kylebarron: Removing `__aenter__` forces your hand and doesn't let you in the future, say, initialize an aiohttp session here.

Issue #909, opened 2026-04-13
re-write xarray reader to remove rio-xarray and usage of indexes


vincentsarago:
Hypothesis: As seen in #902 and #908 we can achieve reading Zarr without the need of spatial indexes. If transform/crs are set we can in theory use a simple `Positional` selection

vincentsarago:
done in #914 as an experimental reader

PR #914, opened 2026-04-15
feat: add experimental geoxarray reader

vincentsarago:
ref #909 

```python
import xarray
from obstore.store import HTTPStore
from zarr.storage import ObjectStore
from affine import Affine
from rasterio.crs import CRS
from matplotlib.pyplot import imshow

from rio_tiler.experimental.xarray import GeoArrayReader

# Create Obstore Store
store = HTTPStore("https://s3.explorer.eopf.copernicus.eu/esa-zarr-sentinel-explorer-fra/tests-output/sentinel-2-l2a/S2B_MSIL2A_20260216T142149_N0512_R096_T25WFV_20260216T165051.zarr")
zarr_store = ObjectStore(store=store, read_only=True)
dt = xarray.open_datatree(
    zarr_store,
    engine="zarr",
    create_default_indexes=False,
    consolidated=True,
)

ds = dt["/measurements/reflectance/r720m"]

attributes = ds.attrs

tr = attributes["spatial:transform"]
transform = Affine(*tr)
crs = CRS.from_user_input(attributes["proj:code"])

with GeoArrayReader(ds["b02"], transform=transform, crs=crs) as dst:
    img = dst.preview()
```

### DIFF
--- a/rio_tiler/experimental/xarray.py
+++ b/rio_tiler/experimental/xarray.py
@@ -138,7 +138,7 @@ class GeoArrayReader(XarrayReader):
     # List of names for the first dimension (e.g time, bands, etc)
     band_names: list[str] | None = attr.ib(default=None)
 
-    _dims: list = attr.ib(init=False)
+    _dims: list = attr.ib(init=False, factory=list)
 
     @options.default
     def _options_default(self):
@@ -178,24 +178,25 @@ class GeoArrayReader(XarrayReader):
             "CRS is required, either as an argument or in `geo-proj` zarr_conventions"
         )
 
-        self._dims = [
-            d
-            for d in self.input.dims
-            if d not in ["x", "y", "spatial_ref", "crs_wkt", "grid_mapping"]
-        ]
-        assert len(self._dims) in [0, 1], "Can't handle >=4D DataArray"
+        if len(self.input.shape) > 2:
+            self._dims = [
+                d
+                for d in self.input.dims
+                if d not in ["x", "y", "spatial_ref", "crs_wkt", "grid_mapping"]
+            ]
 
         # NOTE: Assume shape of array is (bands, height, width) or (height, width)
         # and get height/width from the last 2 dimensions of the shape.
         shape = attributes.get("spatial:shape", self.input.shape)
         self.height, self.width = shape[-2], shape[-1]
 
+        assert len(self.input.shape) in [2, 3], "Can't handle 1D or >=4D DataArray"
         if len(self.input.shape) == 3:
             self.nbands = self.input.shape[0]
         else:
             self.nbands = 1
 
-        self.bounds = self.input.attrs.get("spatial:bounds")
+        self.bounds = attributes.get("spatial:bounds")
         if not self.bounds:
             self.bounds = array_bounds(self.height, self.width, self.transform)
 
@@ -224,7 +225,25 @@ class GeoArrayReader(XarrayReader):
         if self.band_names:
             return self.band_names
 
-        return [f"b{ix + 1}" for ix in range(self.nbands)]
+        if not self._dims:
+            coords_name = [
+                d
+                for d in self.input.coords
+                if d
+                not in [
+                    self.input.rio.x_dim,
+                    self.input.rio.y_dim,
+                    "spatial_ref",
+                    "crs_wkt",
+                    "grid_mapping",
+                ]
+            ]
+            if coords_name:
+                return [str(self.input.coords[coords_name[0]].data)]
+
+            return [self.input.name or "b1"]  # type: ignore
+
+        return [str(band) for d in self._dims for band in self.input[d].values]
 
     @property
     def minzoom(self):
--- a/rio_tiler/experimental/xarray.py
+++ b/rio_tiler/experimental/xarray.py
@@ -1,0 +1,750 @@
+"""Xarray experimental reader."""
+
+from __future__ import annotations
+
+import math
+import warnings
+from typing import Any, cast
+
+import attr
+import numpy
+from affine import Affine
+from morecantile import Tile, TileMatrixSet
+from rasterio.crs import CRS
+from rasterio.transform import array_bounds, rowcol
+from rasterio.warp import calculate_default_transform
+from rasterio.warp import transform as transform_coords
+from rasterio.warp import transform_bounds
+from rasterio.windows import from_bounds as window_from_bounds
+
+from rio_tiler._warp import warp
+from rio_tiler.constants import (
+    MAX_ARRAY_SIZE,
+    WEB_MERCATOR_CRS,
+    WEB_MERCATOR_TMS,
+    WGS84_CRS,
+)
+from rio_tiler.errors import (
+    InvalidGeographicBounds,
+    MaxArraySizeError,
+    PointOutsideBounds,
+    TileOutsideBounds,
+)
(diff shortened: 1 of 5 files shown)


emmanuelmathot:
Good to go. Test here: https://api.explorer.eopf.copernicus.eu/rstaging/collections/sentinel-2-l2a-staging-codecs/items/S2B_MSIL2A_20260503T112109_N0512_R037_T30TTM_20260503T134410/viewer

emmanuelmathot:
**Bug: zarr v3 `fill_value: "NaN"` is not detected as nodata in `GeoArrayReader`, causing NaN pixels to render as 0**

The nodata detection in `GeoArrayReader._read()` (and `info()`) checks:
```python
self.options.get("nodata", self.input.attrs.get("fill_value"))
```

For zarr v3 arrays, the fill value lives in the **array-level metadata** (`fill_value: "NaN"` in `zarr.json`), not in user attributes. When xarray opens such an array it stores this in `da.encoding["_FillValue"]` — not `da.attrs` — so the lookup returns `None`.

With `nodata = None` the masking block is skipped entirely. The raw float32 data already contains NaN values (the `cast_value` codec decodes stored zeros back to NaN), but they are never masked, so they are cast to 0 on render.

For reference, the async `zarr.Reader` handles this correctly with:
```python
nodata = getattr(self.input.metadata, "fill_value", None)
```
which returns the actual `float("nan")` directly from the zarr-python metadata object.

Two possible fixes (combined or separately):

1. **Extend the detection chain** to also probe `self.input.encoding.get("_FillValue")` — where xarray stores the zarr fill value.

2. **Auto-mask NaN for float arrays** as a fallback (mirrors what `da.to_masked_array()` does in `XarrayReader`):
```python
elif numpy.issubdtype(data.dtype, numpy.floating):
    masked_data.mask = numpy.isnan(masked_data.data)
```
This is robust regardless of fill value storage, since zarr v3 always decodes to actual NaN in the float array when `fill_value: "NaN"`.

vincentsarago:
@emmanuelmathot regarding your comment, I'm not quite sure to follow everything. 

I know there is an issue with Nodata because I don't think there is a clear convention about where to store this `nodata` sentinel value. The zarr V3 fill_value metadata describe how the missing chuck should be interpreted but this is not the same as `nodata` within an array. 

Xarray and Zarr python seems to have different implementation on this (see slack thread). 

I will want to avoid to do like rioxarray https://github.com/corteva/rioxarray/blob/67e7b63e77c5ad16f6c97ff122faaa0850470fd0/rioxarray/raster_array.py#L162-L196

emmanuelmathot:
@vincentsarago You're right that zarr v3 `fill_value` and CF-style `nodata` are semantically distinct concepts. But in practice, for this use case they converge:

**The zarr v3 spec defines `fill_value` as the value used to represent uninitialized / missing data.** In Sentinel-2 GeoZarr products, `fill_value: "NaN"` *is* the intended sentinel for "no valid observation here". This is not ambiguous, OpenLayers recently implemented exactly this in [openlayers/openlayers#17355](https://github.com/openlayers/openlayers/pull/17355): read `fill_value` from zarr metadata, use it as nodata, discard those pixels.

**What's wrong with the rioxarray approach** ([rioxarray/raster_array.py#L162–196](https://github.com/corteva/rioxarray/blob/67e7b63e77c5ad16f6c97ff122faaa0850470fd0/rioxarray/raster_array.py#L162-L196)): it walks a long priority chain over many different attribute keys (`_FillValue`, `missing_value`, `nodata`, `nodatavals`, rasterio manager…), trying to reconcile CF, HDF5 and rasterio conventions. It is complex, brittle, and still doesn't robustly handle zarr v3 because it primarily inspects `da.attrs` rather than zarr metadata.

**What xarray actually does today**: when opening a zarr v3 array, xarray reads `fill_value` from the zarr array metadata and surfaces it in `da.encoding["_FillValue"]`. There is an open xarray issue ([pydata/xarray#11332](https://github.com/pydata/xarray/issues/11332)) where @maxrjones is working on making `_FillValue` in zarr attributes accept JSON-native scalars properly. Until that lands, `da.encoding.get("_FillValue")` is the right place to read it which is what the async `zarr.Reader` already does via `self.input.metadata.fill_value`.

**Proposed minimal fix for `_read()`** — no complex convention chain, no rioxarray pattern:

```python
# Read zarr fill_value from xarray encoding (where xarray stores the zarr metadata fill_value)
if nodata is None:
    nodata = self.options.get(
        "nodata",
        self.input.encoding.get("_FillValue")
    )
```

And as a safety net for float arrays (matches `da.to_masked_array()` behavior):

```python
masked_data = numpy.ma.masked_invalid(data)  # masks NaN and Inf for float dtypes
if nodata is not None and not numpy.isnan(nodata):
    masked_data = numpy.ma.masked_equal(masked_data, nodata)
```

The same fix applies to `info()`.

rio_tiler/experimental/xarray.py
```diff
@@ -0,0 +1,750 @@
...
+
+    @property
+    def _nodata_value(self) -> NoData | None:
+        return self.input.attrs.get(
+            "_FillValue",
+            self.input.attrs.get(
+                "missing_value",
+                self.input.attrs.get("fill_value", self.input.attrs.get("nodata")),
+            ),
+        )
```
vincentsarago, line 269: CF style nodata

tests/test_geoxarray.py
```diff
@@ -0,0 +1,487 @@
...
+    dataset = zarr.create_group(store=store)
+    zarrobj = dataset.create_array(
+        "data",
+        shape=(100, 100),
+        chunks=(50, 50),
+        dtype="float32",
+        fill_value=-9999.0,
+        dimension_names=("y", "x"),
+        attributes={
+            "_FillValue": FillValueCoder.encode(-9999.0, numpy.dtype("float32")),
```
vincentsarago, line 539: in this case, xarray will remove `_FillValue` from the attribute and replace those values by `nan` in the array.

We adapted the code to first use `DataArray.to_masked()` which will `mask` those `nan` value but also allow nodata overwriting

vincentsarago:
tests are still failing for python 3.14 

here is what Claude is telling me 


### Root cause analysis:

- _apply_mask (xarray/coding/variables.py:132) calls np.where(condition, decoded_fill_value, data)
- data comes from ZarrArrayWrapper.__getitem__ → zarr v3's sync wrapper → zarr's I/O pipeline
- If zarr v3 returns an array that shares memory with an internal chunk cache buffer (common in chunked I/O), and zarr's background threads evict/reuse that buffer while np.where holds a reference to it, you get a use-after-free crash at the C level
- CPython 3.14 may have changed internal memory management in ways that make this easier to trigger
The xarray fix:
Change _apply_mask line 128 from np.asarray (may return a view) to np.array (always copies), ensuring the data is independent of zarr's buffer before numpy operations:

```python
# xarray/coding/variables.py  
def _apply_mask(data, encoded_fill_values, decoded_fill_value, dtype):
-   data = np.asarray(data, dtype=dtype)
+   data = np.array(data, dtype=dtype)   # always copies; zarr chunks may share memory
    condition = False
    for fv in encoded_fill_values:
        condition |= data == fv
    return np.where(condition, decoded_fill_value, data)
```

vincentsarago:
xarray doesn't support python3.14 officially yet so we can ignore the segfault for now

PR #918, opened 2026-04-22
add async-multibase and async-stac-reader

vincentsarago:


rio_tiler/io/stac.py
```diff
@@ -533,3 +538,213 @@ def _get_asset_info(self, asset: AssetType) -> AssetInfo:  # noqa: C901
...
+                    yield src
+
+
+            async with AsyncSTACReader(input=item, reader=reader) as stac:
+                await stac.tile(...)
+
+
+    """
+
+    input: pystac.Item = attr.ib(converter=_to_pystac_item)
```
vincentsarago: This is slightly different from the current STACReader because it expects a STAC Item not a url

PR #964, opened 2026-07-14
perf(utils): optimize get_array_statistics

manand881:
- hoist masked_coverage.compressed() (was called 3x per band)
- use counts.argmax()/argmin() instead of counts.tolist().index() for majority/minority
- sort once for all percentiles + median via sequence-aware _weighted_quantiles
- derive valid/masked pixel counts from data_comp.size (one pass vs two ma.count calls)
- vectorize categorical histogram with searchsorted instead of python dict loop

~2-3x faster on 1Mpx arrays; results unchanged.

### DIFF
--- a/rio_tiler/utils.py
+++ b/rio_tiler/utils.py
@@ -66,11 +66,12 @@ def _missing_size(w: int | None = None, h: int | None = None):
 def _weighted_quantiles(
     values: NDArray[numpy.floating],
     weights: NDArray[numpy.floating],
-    quantiles: float = 0.5,
-) -> float:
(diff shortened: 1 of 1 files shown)


manand881:
@vincentsarago

manand881:
@vincentsarago addressed pipeline failures

manand881:
Thank you for accepting my contribution. Is there any release date set for this change? My organisation would like to use this change in production.

manand881:
@vincentsarago

vincentsarago:
@manand881 I'll try my best to get it today

PR #969, opened 2026-07-20
Chore/: Stricter-statictics-checks

manand881:
Explicitly cast weighted quantile results to float so statistics for integer rasters retain their previous output types. Add regression assertions for median and percentile values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ tests/fixtures/mask*
 .vscode/settings.json
 
 dev_notebooks/*
+
+.DS_Store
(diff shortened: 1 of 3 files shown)


manand881:
@vincentsarago some conditions in the statistics methods didnt have checks and i added support for it even though my previous commit was not breaking anything. this is just for safety

vincentsarago:
@manand881 do you have other upcoming PR or can I do a release with this ?

manand881:
we can go ahead. @vincentsarago

PR #974, opened 2026-07-28
perf: avoid unnecessary copy and mask work in ImageData.render

manand881:
Skip the always-on array copy on the hot path and only copy when dtype rescale runs, since rescale_image mutates in place. Build the mask only when add_mask is True so unused alpha work is skipped. Encoded bytes and layout stay the same. Tests cover non-mutation across drivers, alpha_mask, colormap, rescale, and add_mask band counts.

@vincentsarago


vincentsarago (approved):
😍

Issue #29, opened 2018-02-27
[Explore] Replace PIL image encoding with rasterio/gdal


vincentsarago:
GDAL/Rasterio is more efficient for image encoding than Pillow. 

Note: Also removing Pillow will make lambda function lighter ;-)

vincentsarago:
```python 
def array_to_img_rasterio(arr, mask, color_map=None, img_format="PNG"):
    """
    """
    if arr.dtype != numpy.uint8:
        arr = arr.astype(numpy.uint8)

    nbands = arr.shape[0] if img_format == "JPEG" else arr.shape[0] + 1
    with MemoryFile() as memfile:
        with memfile.open(driver=img_format,
                          dtype=numpy.uint8,
                          count=nbands,
                          height=arr.shape[1],
                          width=arr.shape[2]) as dst:
            dst.write(arr, indexes=list(range(1, arr.shape[0] + 1)))
            if color_map:
                dst.write_colormap(1, color_map)

            if mask is not None and img_format != "JPEG":
                dst.write(mask.astype(numpy.uint8), indexes=nbands)

        return memfile.read()
```
This could replace https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/utils.py#L506-L574 but sadly I can't make it work with colormap image. If my understanding is correct Rasterio need to close the file for writing the colormap.

vincentsarago:
🤦‍♂️ why try complicated things when Numpy can do it for you 

```python
def array_to_img_rasterio(arr, mask, color_map=None, img_format="PNG"):
    """
    """
    if arr.dtype != numpy.uint8:
        arr = arr.astype(numpy.uint8)

    if color_map is not None:
        arr = numpy.transpose(color_map[arr][0], [2, 0, 1]).astype(numpy.uint8)

    if mask is not None and img_format != "JPEG":
        nbands = arr.shape[0] + 1
    else:
        nbands = arr.shape[0]

    if img_format == "JPEG2000":
        img_format = "JP2OpenJPEG"
    with MemoryFile() as memfile:
        with memfile.open(driver=img_format,
                          dtype=numpy.uint8,
                          count=nbands,
                          height=arr.shape[1],
                          width=arr.shape[2]) as dst:
            dst.write(arr, indexes=list(range(1, arr.shape[0] + 1)))

            if mask is not None and img_format != "JPEG":
                dst.write(mask.astype(numpy.uint8), indexes=nbands)

        return memfile.read()
```

[...]

PR #978, opened 2026-07-29
perf: optional fast uint8 PNG/JPEG/WEBP encode (opt-in)

manand881:
Add an opt-in fast encode path for uint8 PNG/JPEG/WEBP tiles in `rio_tiler.utils.render` and `ImageData.render`, aimed at tile servers (e.g. TiTiler) where the encode step is on the hot path. The default install path is unchanged: GDAL via rasterio MemoryFile, byte-for-byte.

Activation
  - `render(..., fast_encode=True)` / `ImageData.render(..., fast_encode=True)`
  - or env `RIO_TILER_FAST_ENCODE=1` (also `true`/`yes`/`on`, case-insensitive)
  - Explicit kwarg wins over env. Default is off, so `pip install rio-tiler` keeps the current GDAL-only behavior — no ambient switch just because a backend is importable.

Backend selection
  - imagecodecs is the only fast backend wired up here (Pillow tier was dropped during prototyping; imagecodecs covers PNG/JPEG/WEBP uint8).
  - On any backend failure (ImportError or encode error) the code falls back to the existing GDAL path and wraps failures as `InvalidFormat` as before.
  - Non-uint8 dtypes, 4-band JPEG (GDAL CMYK), and any creation option not in the supported set (e.g. WORLDFILE, PROGRESSIVE) bail to GDAL immediately, so unsupported inputs are byte-identical to the off path.

GDAL-aligned defaults (preserved when creation options are omitted)
  - JPEG QUALITY 75, PNG ZLEVEL 6, WEBP QUALITY 75 / LOSSLESS False.
  - `img_profiles` presets are still honored when passed explicitly (e.g. `img_profiles['jpeg']` quality 85, `img_profiles['pngraw']` zlevel 1).
  - WEBP LOSSLESS now interprets GDAL-style strings correctly (`bool('FALSE')` was True in Python; `FALSE`/`NO`/`OFF`/`0` now map to lossy).
  - Out-of-range QUALITY/ZLEVEL are clamped on the fast path instead of raising (GDAL rejects them with InvalidFormat); clamped output matches the output at the clamped in-range value.

Behavior contracts (locked by tests in tests/test_utils.py / tests/test_models.py)
  - Lossless PNG is decode-equal to GDAL (pixels, mask, layout, band count).
  - Lossy JPEG/WEBP keep layout/semantics parity (driver, band count, alpha rules, gray→RGB expand for WEBP) but are not guaranteed byte-identical to GDAL at the same quality.
  - `ImageData.array` is not mutated on render (uint8 or rescale path).
  - `add_mask=False` omits alpha as before.
  - JPEG 4-band stays GDAL CMYK (byte-identical fallback).
  - Concurrent renders from multiple threads are safe.

Estimated speedups (256x256 uint8 tiles, imagecodecs backend vs GDAL path, AMD Ryzen 7 9700X 8c/16t, 60 GiB RAM, Python 3.11.15, mean over repeats):
  - JPEG 3-band:           2.21 ms -> 0.17 ms  (~13x)
  - ImageData JPEG:        2.29 ms -> 0.20 ms  (~11x)
  - PNG 1-band:            1.27 ms -> 1.11 ms  (~1.15x)
  - PNG 3-band+mask z1:    5.99 ms -> 5.65 ms  (~1.06x)
  - PNG 3-band+mask z6:    7.45 ms -> 7.22 ms  (~1.03x)
  - WEBP 3-band+mask:      7.26 ms -> 6.84 ms  (~1.06x)
  - WEBP 1-band:           6.05 ms -> 5.76 ms  (~1.05x)
  JPEG wins big because the fast path skips rasterio MemoryFile + GDAL driver
  setup; PNG/WEBP gains are modest since they remain zlib/codec-bound.

Packaging and CI
  - New optional extra `rio-tiler[fast-encode]` (imagecodecs>=2024.1.1).
  - imagecodecs added to the `dev` dependency group so local/CI can exercise the fast path without the extra.
  - CI runs the default suite (fast_encode off) plus a second pass over tests/test_utils.py and tests/test_models.py with `RIO_TILER_FAST_ENCODE=1` on the latest Python.

Docs
  - README, docs/src/intro.md, docs/src/models.md, and docs/src/advanced/dynamic_tiler.md document the extra, the env/kwarg opt-in, the GDAL-aligned defaults, and the lossy-not-bit-identical caveat.
  - CHANGES.md updated with the full entry and the LOSSLESS string fix.
  
  @vincentsarago

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv sync
+          # dev group includes imagecodecs/pillow so fast_encode tests can run
+          uv sync --group dev
 
       - name: Run pre-commit and mypy
         if: ${{ matrix.python-version == env.LATEST_PY_VERSION }}
@@ -39,7 +40,16 @@ jobs:
           uv run --with mypy --with types-attrs --with types-cachetools mypy -p rio_tiler --ignore-missing-imports
 
       - name: Run tests
+        # Default: fast encode off (GDAL path) unless tests pass fast_encode=True
         run: uv run pytest --cov rio_tiler --cov-report xml --cov-report term-missing -s -vv
+
+      - name: Run render tests with fast encode enabled
+        # Env opt-in path (imagecodecs/Pillow) for uint8 PNG/JPEG/WEBP
+        if: ${{ matrix.python-version == env.LATEST_PY_VERSION }}
+        env:
+          RIO_TILER_FAST_ENCODE: "1"
+        run: |
+          uv run pytest tests/test_utils.py tests/test_models.py -q --tb=short
 
       - name: Upload Results
         if: ${{ matrix.python-version == env.LATEST_PY_VERSION }}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 
 # Unreleased
 
+* perf: optional fast uint8 PNG/JPEG/WEBP encode path in `utils.render` / `ImageData.render` (author @manand881)
+    * **Opt-in only** (default remains GDAL): `fast_encode=True` or env `RIO_TILER_FAST_ENCODE=1` (`true`/`yes`/`on`). Explicit kwarg overrides env.
+    * Backends when enabled: **imagecodecs → GDAL** fallback. Install with `pip install "rio-tiler[fast-encode]"` (also in the `dev` dependency group).
+    * Omitting creation options uses **GDAL driver defaults** (JPEG quality **75**, PNG zlevel **6**, WEBP quality **75**, lossless false). `img_profiles` still apply when passed.
+    * Unsupported creation options (anything other than the mapped quality/zlevel/lossless keys) fall back to GDAL.
+    * Lossless **PNG** is decode-equal to GDAL (pixels/mask/layout). Lossy **JPEG/WEBP** may not be bit-identical to GDAL at the same quality.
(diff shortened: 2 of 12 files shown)


manand881:
Hi @vincentsarago ,
Quick question on the fast-encode work in the PR. Right now imagecodecs is optional (rio-tiler[fast-encode]) and the fast path is opt-in via fast_encode=True / RIO_TILER_FAST_ENCODE=1. I'm wondering whether we should just make imagecodecs a regular dependency and flip the default on.
The main reason: the JPEG win is big. On 256×256 uint8 tiles I'm measuring ~13x for utils.render JPEG (2.21 ms → 0.17 ms) and ~11x for ImageData.render JPEG (2.29 ms → 0.20 ms) vs the GDAL MemoryFile path. For tile servers like TiTiler that's the hot path, and most tiles are JPEG or PNG. PNG/WEBP are more modest (1.03–1.15x) since they stay zlib/codec-bound, but JPEG alone carries the win.

manand881:
@vincentsarago Once this Branch is finalized can we release it?

vincentsarago:
dear @manand881 
I appreciate your contribution 🙏 

This PR is pretty big and adds new dependency which I usually try to avoid. For big `features` like this it's always better to open an issue first 

To me the performance gain seems (IMO) negligible in comparison of other bottleneck in rio-tiler 

https://github.com/cogeotiff/rio-tiler/issues/29
https://github.com/vincentsarago/benchmark-img-gdal

if you really need this in titiler you can customize your application by providing a custom `render_func` https://github.com/developmentseed/titiler/blob/2b547a2357fd675d0ddc2f9ed70fb2f4a4c77188/src/titiler/core/titiler/core/factory.py#L331

PR #755, opened 2024-10-24
expand xarray capabilities

vincentsarago:
This PR does:
- adds `preview()` and `statistics()` method to the XarrayReader
- add output size (max_size | width, height) options for `preview`, `part` and `feature` methods

⚠️ Preview and Statistics methods will need to first load the full resolution data in memory and should be used with care! 

cc @abarciauskas-bgse @maxrjones @hrodmn

### To Do

- [x] update docs

Issue #754, opened 2024-10-23
InvalidGeographicBounds exception raised for datasets covering whole WGS84 extent


noahgolmant:
Right now, if a dataset has extent equal to (-180, -90, 180, 90), then `XarrayReader` will [throw](https://github.com/cogeotiff/rio-tiler/blob/abec15162c1d8cf9dd67cb5945d8532e2b6f562c/rio_tiler/io/xarray.py#L79) an `InvalidGeographicBounds` exception. This is because `self.rio.bounds` actually applies a single pixel buffer to the internal dataset bounds based on the resolution of the dataset (see [here](https://github.com/corteva/rioxarray/blob/f3ad955dd59dfdd1a696ac9d4a4b5be1418c5994/rioxarray/rioxarray.py#L1079)).

Is this the desired behavior? If not, the fix should be relatively simple since we can apply a similar buffer to the bounds in that check.

vincentsarago:
@noahgolmant feel free to start a PR 🙏

PR #994, opened 2026-09-07
fix: tolerate float residue in the WGS84 bounds check

javandres:
## Problem

`XarrayReader` and `ZarrReader` allow half a cell of overshoot beyond
`(-180, -90, 180, 90)`, so that a global grid whose outer cells are centred on
the poles and the antimeridian is accepted. That allowance came in with #755,
fixing #754.

The comparison is exact, and coordinate arrays are not. Coordinates built with
`numpy.arange` — which is how a lot of published data is written — accumulate a
float residue, and the allowance then misses by that residue:

```python
import numpy, xarray, rioxarray

y = numpy.arange(90, -90.4, -0.4)     # 0.4 degree global grid, cells centred on the poles
x = numpy.arange(-180, 180, 0.4)
da = xarray.DataArray(
    numpy.zeros((len(y), len(x)), dtype="float32"),
    dims=("y", "x"), coords={"y": y, "x": x},
)

XarrayReader(da)
# rio_tiler.errors.InvalidGeographicBounds: Invalid geographic bounds:
# (-180.20000000000002, -90.20000000000256, 179.80000000000516, 90.2).
```

`bounds[1] + yres / 2` is `-90.00000000000256` — the dataset is rejected by
2.6e-12 degrees, which is float noise, not data.

Measured residue on `south + yres / 2`, global grids built this way:

| resolution | rows | residue |
|-----------:|-----:|--------:|
| 1.0°, 0.5°, 0.25° | 181–721 | 0 |
| 0.4° | 451 | 2.6e-12 |
| 0.1°, 0.05° | 1801–3601 | 1.0e-11 |
| 0.01° | 18001 | 9.2e-11 |
| 0.001° | 180001 | 8.6e-10 |

We hit this on the CAMS global atmospheric composition forecast (ECMWF), a 0.4°
store whose south edge lands on `-90.20000000000041`. Served through
`titiler.xarray`, **every tile answered HTTP 500**, and nothing in the message
suggests the data is fine and the arithmetic is not.

## Fix

Allow a thousandth of a cell of slack on each of the four comparisons, in both
readers that carry the allowance (`io/xarray.py` and `experimental/_zarr.py`).

A relative tolerance rather than an absolute one, because as the table shows the
residue grows with the number of steps: an absolute `1e-9` is comfortable at 0.4°
and already marginal at 0.001°. Scaling with the cell holds at any resolution.

A thousandth of a cell is far below any real overshoot — a genuinely misplaced
dataset is out by whole cells, as the existing `test_xarray_reader_invalid_bounds_crs`
cases are, and those still raise.

Happy to switch to an absolute epsilon, or to `math.isclose` semantics, if you
prefer one of those.

## Not touched

`experimental/xarray.py::GeoArrayReader` carries a similar check, but its bounds
come from an explicit affine transform / GeoZarr `spatial` conventions rather
than from accumulated coordinates, so there is no residue of this kind to absorb
and the strict comparison there looks correct. Left alone.

## Tests

- `tests/test_io_xarray.py::test_xarray_reader_global_bounds_float_precision`
- `tests/test_io_zarr.py::test_dataset_reader_global_bounds_float_precision`

Both fail on `main` with `InvalidGeographicBounds` and pass with the change.
Full suite: 459 → 461 passed, same 12 xfailed / 6 xpassed, no regressions.
`pre-commit` clean.


vincentsarago:
thanks for opening the PR @javandres and adding verbose explanation 🙏 

I'm always supper hesitant to add fixes to overcome issues with bad datas. Clearly there are issues with how xarray/zarr describe their coordinates and people are missing specification to follow (hopefully GeoZARR fixes this). 

> Happy to switch to an absolute epsilon, or to math.isclose semantics, if you prefer one of those.

I think this would be better

javandres:
@vincentsarago Switched to `math.isclose`.

On the "bad data" point, worth separating two things. The coordinates here are
fine: a 0.4° global grid centred on the poles *should* report bounds of -90.2,
and #755 accepts exactly that. What is off is the check, which compares floats
with a strict `<`. `numpy.arange(90, -90.4, -0.4)` puts the south cell centre on
-90.00000000000256 rather than -90.0, and that 2.6e-12 is the entire reason the
dataset is refused. No coordinate convention fixes that — only the comparison can.

Measuring in order to answer you turned up a hole in my own version, so the
question was worth asking. The residue is float noise on numbers of magnitude 90
and 180, so it scales with the *coordinate*, not with the cell — it grows as the
grid gets finer while a per-cell tolerance shrinks:

| grid | south centre past the pole | thousandth of a cell | absolute 1e-9 | `math.isclose` |
|---|---|---|---|---|
| 0.4° | 2.6e-12 | ok | ok | ok |
| 0.001° | 8.6e-10 | ok | ok | ok |
| 0.0001° | 6.0e-09 | ok | **rejects** | ok |
| 2e-05° | 5.7e-08 | **rejects** | **rejects** | ok |

So both options I offered were wrong: an absolute epsilon has to be picked for
the finest grid it will ever see, and my per-cell one claimed to hold "at any
resolution" and does not. `math.isclose` is already denominated in the right
unit. Its default `rel_tol=1e-9` allows 9e-08° at the poles and 1.8e-07° at the
antimeridian — one to two centimetres, inside a half-cell allowance that is
22 km on the grid that prompted this. `test_xarray_reader_invalid_bounds_crs` is
out by 170° and still raises.

Two module-level helpers in `io/xarray.py` (`_below` / `_above`), imported by
`experimental/_zarr.py`, which already imports `XarrayReader` from there. New
test at 0.0001°, the case that separates `isclose` from a fixed epsilon.
462 passed / 12 xfailed / 6 xpassed, pre-commit clean.

vincentsarago:
@javandres can you check if I have write access to your branch so I can make some changes?

javandres:
> @javandres can you check if I have write access to your branch so I can make some changes?

Yes — "Allow edits by maintainers" is on, and the fork is under my personal
account, so you should be able to push to
javandres:fix/geographic-bounds-float-tolerance directly. Go ahead.

I won't touch the branch until you say you're done, so nothing of yours gets
force-pushed over.

vincentsarago:
@javandres I've pushed some change. Basically just create a `check_bounds` function with the two new functions you added.

Let me know if this is fine 👍

javandres:
@vincentsarago Looks good — better than what I had, thank you!

vincentsarago (approved):

